Issue #166, opened 2026-04-14
BUG-028: merge_segments silently returns Ok when fewer than two requested segment IDs actually exist, masking typos and half-committed requests


davidkelley:
# BUG-028: merge_segments silently returns Ok when fewer than two requested segment IDs actually exist, masking typos and half-committed requests

**Category:** Reliability
**Severity:** Low
**File(s):** searchlite-core/src/index/mod.rs
**Line(s):** 259-286

## Description
`Index::merge_segments(segment_ids, write_key)` validates the requested segment IDs in two stages:

1. `if merge_metas.len() < 2 { return Ok(()); }` (line 275-278) — silent no-op.
2. `if merge_metas.len() != merge_set.len() { bail!(...) }` (line 280-286) — error on mismatch.

The early return at step (1) fires before the mismatch check, so the two error paths are unreachable in combination:

- Caller requests `["A", "B"]`, both exist → proceed (OK).
- Caller requests `["A", "B"]`, only `A` exists → `merge_metas.len() == 1 < 2` → **silent Ok**. The caller is told the merge succeeded; it never ran.
- Caller requests `["A", "B", "C"]`, only `A` and `B` exist → `merge_metas.len() == 2 < 3`, fails the mismatch check → **loud error**.
- Caller requests `["A"]` → `merge_metas.len() == 1 < 2` → silent Ok. Trivial no-op, consistent with semantics.

The interesting asymmetry is the 2nd case: requesting a merge of two segments where one of the IDs is a typo (or refers to a just-deleted segment) is silently ignored rather than surfacing a "segment not found" error. The caller's retry loop has no way to distinguish between "merge succeeded and produced nothing to do" and "merge silently dropped because one of my IDs was wrong." In long-running agents orchestrating manual merges (e.g., an external controller computing merge batches), this makes errors in the merge-plan selection invisible.

## Reproduction / Evidence
```rust
// index/mod.rs:275-286
if merge_metas.len() < 2 {
  return Ok(());          // <-- fires before the missing-id check
}
if merge_metas.len() != merge_set.len() {
  bail!(
    "some segment IDs not found in manifest (requested {}, found {})",
    merge_set.len(),
    merge_metas.len()
  );
}
```

Reproducer:
```rust
let idx = Index::create(path, schema, opts).unwrap();
// ... write & commit one segment "S" ...
idx.merge_segments(&["S".into(), "nonexistent".into()], None)?;  // returns Ok, no merge happened
```

## Expected Behaviour
[...]

PR #184, opened 2026-04-14
fix(merge): surface missing segment IDs in merge_segments (BUG-028)

davidkelley:
## Summary

Fixes BUG-028 (#166). `Index::merge_segments` checked `merge_metas.len() < 2` before verifying that every requested ID resolved to a real segment. When a caller passed two IDs and only one existed, the filtered set shrank below the `< 2` threshold and the function silently returned `Ok(())`, hiding typos and half-committed requests from callers.

## Changes

- Reorder the two validation steps in `searchlite-core/src/index/mod.rs` so ID-existence is verified *before* the "nothing to merge" short-circuit.
- Expand the error message to list the specific missing segment IDs (in addition to the requested/found counts) so operators and orchestrators can diagnose typos immediately.
- Keep the existing no-op semantics for the well-defined cases:
  - empty input (`segment_ids.is_empty()`): `Ok(())`
  - a single valid segment ID: `Ok(())` (nothing to merge, but all IDs resolved)

## Test plan

- [x] New regression test `merge_segments_errors_when_any_id_missing` covers:
  - `[real, typo]` → error naming the typo
  - `[missing, missing]` → error naming both
  - `[]` and `[real]` → still `Ok(())`, manifest unchanged
- [x] Existing `merge_segments_preserves_search` and `commit_with_merge_reduces_segments` still pass (the internal merge planner only feeds valid IDs, so it's unaffected).
- [x] `cargo test -p searchlite-core --lib` — 174 passed.

### DIFF
--- a/searchlite-core/src/index/mod.rs
+++ b/searchlite-core/src/index/mod.rs
@@ -272,17 +272,28 @@ impl Index {
       .iter()
       .filter(|s| merge_set.contains(s.id.as_str()))
       .collect();
-    if merge_metas.len() < 2 {
-      // Nothing useful to merge.
-      return Ok(());
-    }
-    // Verify all unique requested segment IDs exist.
+    // Verify all unique requested segment IDs exist before deciding there is
+    // nothing to merge, so typo'd or half-committed IDs surface as errors
+    // rather than being silently swallowed by the `< 2` early return below.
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

---

Issue #142, opened 2026-04-14
BUG-004: Unchecked u32 subtraction when writing position deltas


davidkelley:
# BUG-004: Unchecked u32 subtraction when writing position deltas

**Category:** Reliability
**Severity:** Medium
**File(s):** searchlite-core/src/index/postings.rs
**Line(s):** 119-125

## Description
`PostingsWriter::write_term` assumes that the `positions` vector for each `PostingEntry` is strictly monotonically non-decreasing, and computes gap deltas with an unchecked subtraction:
```rust
let mut prev = 0;
for pos in p.positions.iter().copied() {
  let delta = pos - prev;     // panics on underflow in debug; wraps in release
  write_u32_var(delta, &mut buf);
  prev = pos;
}
```
The monotonicity invariant is only upheld by convention — `InvertedIndexBuilder::add_term` appends positions in the order tokens are emitted, but *any* caller can construct a `PostingEntry` directly (the field is `pub`) with out-of-order positions. A single out-of-order position produces:
- **Debug:** attempt to subtract with overflow panic, crashing the indexer.
- **Release:** wrapped delta is serialised. On decode this sum re-wraps and produces garbage position data (see BUG-003).

Note that `PostingEntry.positions` being public (postings.rs:19) makes it trivial for external crate code (`searchlite-http`, `searchlite-node`) to construct such an instance.

## Reproduction / Evidence
```rust
use smallvec::smallvec;
let mut file = std::io::Cursor::new(Vec::new());
let postings = vec![PostingEntry {
  doc_id: 0,
  term_freq: 2,
  positions: smallvec![5, 3],  // not monotonic
}];
let mut writer = PostingsWriter::new(&mut file, true);
writer.write_term(&postings).unwrap();  // panics in debug build
```

## Expected Behaviour
Either (a) defensively sort/validate positions before encoding, or (b) return an `Err` that points at the malformed entry. The writer must not silently produce an un-decodable segment and must not panic on malformed user input.

## Actual Behaviour
Panic in debug; silent production of a corrupt segment in release.

## Suggested Fix
Use `checked_sub` and reject non-monotonic input:
```rust
let mut prev = 0u32;
for pos in p.positions.iter().copied() {
  let delta = pos.checked_sub(prev)
    .ok_or_else(|| anyhow::anyhow!(
      "positions must be non-decreasing (doc {} pos {} after {})",
      p.doc_id, pos, prev
    ))?;
  write_u32_var(delta, &mut buf);
[...]

PR #185, opened 2026-04-14
fix(postings): reject non-monotonic positions in write_term (BUG-004)

davidkelley:
## Summary

Fixes BUG-004 (#142).

`PostingsWriter::write_term` delta-encodes each position with an unchecked `u32` subtraction (`pos - prev`). `PostingEntry` is a public type and its `positions` field is directly constructible by callers (including other workspace crates), so the monotonicity invariant is upheld only by convention. A caller that supplies non-monotonic positions hits:

- **Debug builds:** an arithmetic-overflow panic in the writer, crashing the indexer.
- **Release builds:** the subtraction wraps and a malformed delta is written to disk — the segment then decodes into garbage positions and surfaces the complementary BUG-003 overflow on the read path (now already guarded, but still produces a load error rather than accurate positional data).

Swap the unchecked subtraction for `checked_sub` with a descriptive error that identifies the offending `doc_id` and position pair so the writer fails loudly instead of producing a segment that cannot be round-tripped. The legitimate call path through `InvertedIndexBuilder::add_term` already appends positions in token order, so well-formed builds are unaffected.

## Changes

- `searchlite-core/src/index/postings.rs`
  - Replace `let delta = pos - prev;` in `PostingsWriter::write_term` with `pos.checked_sub(prev)` and return an `anyhow::Error` naming the offending `doc_id`, `pos`, and previous position.
  - Add regression tests:
    - `write_term_rejects_non_monotonic_positions` — `[5, 3]` is rejected with a `non-decreasing` error.
    - `write_term_rejects_backward_step_mid_list` — `[0, 10, 4]` is rejected (guards against an off-by-one that only checks the first pair).
    - `write_term_allows_equal_consecutive_positions` — `[2, 2, 5]` remains valid and round-trips through `PostingsReader` intact (delta == 0 must still be accepted).

## Test plan

- [x] `cargo test -p searchlite-core --lib index::postings` — 8 passed, including the three new tests
- [x] `cargo test --workspace --lib` — all existing tests still green (179 in `searchlite-core`, 44 in `searchlite-http`, 7 in `searchlite-ffi`)

### DIFF
--- a/searchlite-core/src/index/postings.rs
+++ b/searchlite-core/src/index/postings.rs
@@ -117,9 +117,23 @@ impl<'a, W: Write + Seek + ?Sized> PostingsWriter<'a, W> {
       write_u32_var(p.term_freq, &mut buf);
       if self.keep_positions {
         write_u32_var(p.positions.len() as u32, &mut buf);
-        let mut prev = 0;
+        let mut prev = 0u32;
         for pos in p.positions.iter().copied() {
-          let delta = pos - prev;
+          // Positions must be non-decreasing so that delta encoding is
+          // reversible. `PostingEntry` is a public type and its `positions`
+          // field is directly constructible by external callers, so we
+          // cannot rely on convention — reject out-of-order input here
+          // rather than producing a segment that silently fails to decode
+          // (debug: subtract-overflow panic; release: wrapped delta that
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

---

Issue #140, opened 2026-04-14
BUG-002: varint read_u64 has no shift-overflow protection


davidkelley:
# BUG-002: varint read_u64 has no shift-overflow protection

**Category:** Reliability
**Severity:** High
**File(s):** searchlite-core/src/util/varint.rs
**Line(s):** 17-29

## Description
`read_u64` decodes a LEB128 varint from a `&[u8]` without bounding the cumulative `shift`. A malformed (or maliciously crafted) buffer where every byte has the continuation bit set drives `shift` past 63. The expression `(b & 0x7F) as u64 << shift` with `shift >= 64` is defined by Rust as a panic in debug builds and an unspecified (wrapping) value in release builds. Either outcome is wrong for a function that is supposed to return `Err` on bad input.

The companion `read_u32_var` (lines 31-49) correctly guards with `if shift > 28 { return Err(..) }`; `read_u64` has no such guard.

## Reproduction / Evidence
```rust
// varint.rs:17-29
pub fn read_u64(buf: &[u8]) -> Result<(u64, usize)> {
  let mut shift = 0u32;
  let mut value = 0u64;
  for (i, b) in buf.iter().enumerate() {
    let part = (b & 0x7F) as u64;
    value |= part << shift;          // <-- UB / panic once shift >= 64
    if b & 0x80 == 0 {
      return Ok((value, i + 1));
    }
    shift += 7;
  }
  Err(anyhow!("unterminated varint"))
}
```

A 20-byte buffer of `0xFF` bytes reaches `shift = 70` on the 11th iteration and panics in debug / silently corrupts the value in release. This input can come from any on-disk varint whose stream cursor has drifted (see BUG-001 for one realistic path that produces such drift).

## Expected Behaviour
Return `Err("varint too long")` once the shift would exceed 63 for a `u64`.

## Actual Behaviour
Panic in debug; in release, upper bits silently wrap and the function returns an apparently valid `(u64, usize)` that is garbage.

## Suggested Fix
Mirror `read_u32_var`:
```rust
pub fn read_u64(buf: &[u8]) -> Result<(u64, usize)> {
  let mut shift = 0u32;
  let mut value = 0u64;
  for (i, b) in buf.iter().enumerate() {
    let part = (b & 0x7F) as u64;
    value |= part << shift;
    if b & 0x80 == 0 {
      return Ok((value, i + 1));
    }
    shift += 7;
    if shift > 63 {
      return Err(anyhow!("varint too long"));
    }
[...]

PR #201, opened 2026-04-15
fix(varint): guard read_u32_var against final-byte shift overflow (BUG-198)

### LINKED ISSUE #198: fix(varint): read_u32_var has no shift-overflow protection — silently truncates u32-overflowing varints to bogus u32 values
## Summary
`read_u32_var` in `searchlite-core/src/util/varint.rs` does not guard
against shift overflow at the final possible byte (shift == 28). A
maliciously crafted or corrupt varint whose 5th byte has any bit
above bit 3 set (`b & 0x70 != 0`) silently decodes to an arbitrary
truncated u32 value instead of being rejected as overflow. This is the
exact same class of bug as BUG-002 (#140), which was fixed for
`read_u64` in commit 017037d but the sibling `read_u32_var` in the same
file was overlooked.

## Affected code
- [searchlite-core/src/util/varint.rs:41](searchlite-core/src/util/varint.rs#L41) — `read_u32_var` is missing the analog of the `shift == 63 && part > 1` guard that `read_u64` got
- [searchlite-core/src/util/varint.rs:50](searchlite-core/src/util/varint.rs#L50) — the unguarded `value |= ((b & 0x7F) as u32) << shift` is where the high bits silently disappear when `shift == 28`
- Reachable from corrupt segment data via:
  - [searchlite-core/src/index/postings.rs:223](searchlite-core/src/index/postings.rs#L223) — `doc_id`
  - [searchlite-core/src/index/postings.rs:224](searchlite-core/src/index/postings.rs#L224) — `term_freq`
  - [searchlite-core/src/index/postings.rs:227](searchlite-core/src/index/postings.rs#L227) — position `count`
  - [searchlite-core/src/index/postings.rs:232](searchlite-core/src/index/postings.rs#L232) — position delta

## Reproduction / evidence
Failing test sketch (run against current `searchlite-core::util::varint`):

```rust
#[test]
fn read_u32_var_rejects_final_byte_overflow() {
    use std::io::Cursor;
    // 5-byte varint: four continuation bytes (value bits all zero) + final
    // byte 0x10. The final byte's bit 4 sits at position 32 once shifted,
    // i.e. one bit above u32::MAX. The decoded value is exactly 2^32 and
    // does NOT fit in a u32. read_u32_var must error, but today it returns
    // Ok(0) — the high bit is silently dropped by the u32 left shift.
    let buf = vec![0x80, 0x80, 0x80, 0x80, 0x10];
    let mut cur = Cursor::new(buf);
    let err = read_u32_var(&mut cur)
        .expect_err(\"u32-overflowing varint must be rejected\");
    assert!(err.to_string().contains(\"overflows u32\"));

    // Same shape, even more value bits set in the final byte (bits 4..6).
    let buf = vec![0x80, 0x80, 0x80, 0x80, 0x70];
    let mut cur = Cursor::new(buf);
    let err = read_u32_var(&mut cur)
        .expect_err(\"u32-overflowing varint must be rejected\");
    assert!(err.to_string().contains(\"overflows u32\"));
}
```

Today both calls return `Ok(0)`. Trace for `[0x80, 0x80, 0x80, 0x80, 0x10]`:

| iter | shift | byte | `(b & 0x7F) as u32 << shift`                | `value` after |
|------|------:|------|---------------------------------------------|---------------|
| 1    | 0     | 0x80 | `0x00 << 0  = 0x0`                          | `0x0`         |
| 2    | 7     | 0x80 | `0x00 << 7  = 0x0`                          | `0x0`         |
| 3    | 14    | 0x80 | `0x00 << 14 = 0x0`                          | `0x0`         |
| 4    | 21    | 0x80 | `0x00 << 21 = 0x0`                          | `0x0`         |
| 5    | 28    | 0x10 | `0x10 << 28 = 0x1_0000_0000`, truncated to `u32` → `0x0` | `0x0` |

The varint represents the value `2^32`, which is one above `u32::MAX`,
so the correct response is to reject it. Instead the function returns
`Ok(0)`. With `0x70` in the final byte the value is `0x7_0000_0000_0`
(~30 billion) and the truncated result is again `0`.

## Expected behavior
`read_u32_var` should reject any varint whose decoded value cannot fit
in a `u32`, returning an `Err` with a message like `\"varint overflows u32\"`,
matching the behavior `read_u64` got in BUG-002 (#140).

## Actual behavior
The function silently returns `Ok` with the high bits dropped by the
u32 left shift. Corrupt postings data can therefore decode without
error to a doc_id / term_freq / position-count / position-delta that
bears no relation to what was on disk, and the caller has no way to
detect the corruption at this layer.

## Root cause (hypothesis)
BUG-002 (#140) was fixed in commit 017037d only for `read_u64` —
the patch added both the `shift > 63` overlong guard *and* the
`shift == 63 && part > 1` final-byte overflow guard. The function
`read_u32_var` immediately below it (lines 41–59) has the same shape
but received neither guard. The `shift > 28` overlong guard is in
place, but the analog of the `shift == 63 && part > 1` final-byte
check (here it would be `shift == 28 && (b & 0x7F) > 0x0F`) is
missing, so the loop accepts the byte and silently truncates.

For honest writers this is unreachable because `write_u32_var` only
encodes values in `0..=u32::MAX`, whose 5th byte is always
`<= 0x0F`. The bug only matters for corrupt or maliciously crafted
segment files, but the same was true of BUG-002 — the project's
position is evidently to reject such inputs rather than silently
mis-decode them.

## Suggested fix
Mirror the `read_u64` patch:

```rust
pub fn read_u32_var<R: Read>(r: &mut R) -> Result<u32> {
  let mut shift = 0u32;
  let mut value = 0u32;
  loop {
    let mut byte = [0u8; 1];
    if let Err(e) = r.read_exact(&mut byte) {
      return Err(anyhow!(e));
    }
    let b = byte[0];
    let part = (b & 0x7F) as u32;
    // At shift == 28 only the low 4 bits of `part` fit in a u32; any
    // higher bit would be silently truncated by the shift. Reject such
    // inputs as overflowing u32 (mirrors the shift == 63 guard in
    // read_u64; see BUG-002 / #140).
    if shift == 28 && part > 0x0F {
      return Err(anyhow!(\"varint overflows u32\"));
    }
    value |= part << shift;
    if b & 0x80 == 0 {
      return Ok(value);
    }
    shift += 7;
    if shift > 28 {
      return Err(anyhow!(\"varint too long\"));
    }
  }
}
```

## Environment
- Commit: f24b62e
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes #198.

`read_u32_var` in `searchlite-core/src/util/varint.rs` did not guard against shift overflow at the final possible byte (`shift == 28`). A varint whose 5th byte had any bit above bit 3 set (`part > 0x0F`) silently decoded to an arbitrary truncated `u32` value instead of being rejected as overflow. This is the same class of bug as BUG-002 (#140), which was fixed for `read_u64` but the sibling `read_u32_var` in the same file was overlooked.

Honest writers encode values in `0..=u32::MAX`, whose 5th byte is always `<= 0x0F`, so the bug is unreachable for well-formed input. Corrupt or maliciously-crafted segment data, however, could previously smuggle truncated `doc_id` / `term_freq` / position-count / position-delta values past the varint layer. Callers reached through `searchlite-core/src/index/postings.rs:223-232` had no way to detect the corruption at this layer.

## Changes

- `searchlite-core/src/util/varint.rs`
  - `read_u32_var`: add the `shift == 28 && part > 0x0F` guard, returning `"varint overflows u32"`. Directly mirrors the `shift == 63 && part > 1` guard `read_u64` got in BUG-002.
  - Inline comment pointing at BUG-002 / #140 so the parallel with `read_u64` stays obvious the next time someone audits this file.

No on-disk format change, no caller-visible API change, no change to `write_u32_var` / `write_u64` / `read_u64`.

## Regression tests

Four new tests under the existing `#[cfg(test)] mod tests`:

- `read_u32_var_roundtrip_boundary_values` — round-trips `0`, `1`, `127`, `128`, `16383`, `16384`, `u32::MAX - 1`, `u32::MAX` through `write_u32_var` + `read_u32_var` end-to-end. Pins that the stricter guard does not regress honest inputs, including the boundary case where the final byte is exactly `0x0F`.
- `read_u32_var_rejects_final_byte_overflow` — the exact two repro buffers from the issue (`[0x80, 0x80, 0x80, 0x80, 0x10]` → decodes to `2^32` pre-fix, and `[0x80, 0x80, 0x80, 0x80, 0x70]` → decodes to `~30 billion` pre-fix, both truncated to `0` under a `u32` shift). Asserts both now error with exactly `"varint overflows u32"`.
- `read_u32_var_rejects_overlong_varint` — 10 continuation-only bytes must return `"varint too long"`, pinning the pre-existing `shift > 28` overlong guard so the new check doesn't accidentally change that error shape.
- `read_u32_var_accepts_max_length_valid_varint` — asserts `u32::MAX` still encodes to exactly 5 bytes with `0x0F` in the final byte and round-trips unchanged, so the new guard's lower bound (`> 0x0F`) is correct rather than over-aggressive.

## Test plan

- [x] `cargo test -p searchlite-core --lib util::varint` — 11 passed (4 new + 7 pre-existing)
- [x] `cargo test -p searchlite-core --lib` — 202 passed, 0 failed
- [x] `cargo test --workspace --lib --exclude searchlite-wasm` — all green

Refs #198

### DIFF
--- a/searchlite-core/src/util/varint.rs
+++ b/searchlite-core/src/util/varint.rs
@@ -47,7 +47,16 @@ pub fn read_u32_var<R: Read>(r: &mut R) -> Result<u32> {
       return Err(anyhow!(e));
     }
     let b = byte[0];
-    value |= ((b & 0x7F) as u32) << shift;
+    let part = (b & 0x7F) as u32;
+    // On the final possible byte (shift == 28) only the low 4 bits of
+    // `part` fit in a u32; any higher bit would be silently truncated by
+    // the shift. Reject such inputs as overflowing u32 rather than
+    // decoding to a lossy value (mirrors the shift == 63 guard in
+    // read_u64; see BUG-002 / #140).
+    if shift == 28 && part > 0x0F {
+      return Err(anyhow!("varint overflows u32"));
+    }
+    value |= part << shift;
     if b & 0x80 == 0 {
       return Ok(value);
     }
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

davidkelley:
@codex review

---

Issue #150, opened 2026-04-14
BUG-012: Fast-field deserialization sizes Vec::with_capacity from untrusted file counts — OOM / integer overflow on corrupt segment


davidkelley:
# BUG-012: Fast-field deserialization sizes Vec::with_capacity from untrusted file counts — OOM / integer overflow on corrupt segment

**Category:** Reliability
**Severity:** High
**File(s):** searchlite-core/src/index/fastfields.rs
**Line(s):** 1257-1302, 1323-1368, 1371-1381, 1397-1409

## Description
Every list / nested / string-dict variant in the fast-field column reader derives a `Vec::with_capacity(N)` or loop bound from a `u32` read directly out of the mmap'd segment file, with no sanity check against the remaining buffer length. Examples:

- I64List / F64List: `total_vals = *offsets.last().unwrap_or(&0) as usize` → `Vec::with_capacity(total_vals)` (1261-1262, 1327-1328).
- I64Nested / F64Nested: two layers of unbounded counts — `total_objects = *doc_offsets.last() as usize`, then `Vec::with_capacity(total_objects + 1)` (1282-1283, 1348-1349). `total_objects + 1` is also an un-checked `usize` add that will panic in debug if `total_objects == usize::MAX`.
- Str / StrList dict: `dict_len = read_u32(...) as usize` → `Vec::with_capacity(dict_len)` (1372, 1398).

On 64-bit, a hostile (or merely truncated-and-garbage) file with a u32 count near `u32::MAX` causes `Vec::with_capacity` to request ≈4G elements × the element stride (e.g., 32 GiB for `[i64; 4G]`), which aborts the process with an allocator failure. `doc_len + 1` at 1257/1278/1323/1344/1408 has the same overflow-panic issue (`doc_len` is `usize` and can be crafted to `usize::MAX` via the manifest).

This shows up as a denial-of-service on any system that loads segments from untrusted storage (S3 with a compromised bucket, a replica recovering from a poisoned peer, or even a local filesystem with silent corruption). The per-element read loops *do* check `cursor + 8 > data.len()` but only *inside* the loop — the capacity allocation happens before that check fires.

## Reproduction / Evidence
```rust
// fastfields.rs:1277-1286
FieldType::I64Nested => {
  let mut doc_offsets = Vec::with_capacity(doc_len + 1);
  for _ in 0..doc_len + 1 {
    doc_offsets.push(read_u32(&mut cursor, data)?);
  }
[...]

PR #208, opened 2026-04-15
fix(terms): bound Vec capacity against untrusted u64 term_count (BUG-207)

### LINKED ISSUE #207: fix(terms): read_terms sizes Vec::with_capacity from untrusted u64 term_count — OOM / process abort on corrupt terms file
## Summary

`read_terms` derives `Vec::with_capacity` directly from a `u64` `term_count` value read out of the terms file with no validation against the bytes still available in the file. The 8-byte `term_count` header lives **outside** the file's internal CRC (the CRC is computed only over the per-term payload), so a single-byte flip in those eight bytes — or a tampered/legacy segment whose outer manifest checksum is consistent — commits a multi-gigabyte allocation before the per-term read loop ever discovers the file is too short. This is the same class of bug that BUG-012 (#150) fixed inside `read_fields` and that #205 still has open against `PostingsReader::read_at`; `terms.rs` was not covered by either change.

## Affected code

- [searchlite-core/src/index/terms.rs:31-33](searchlite-core/src/index/terms.rs#L31-L33) — `term_count` is read from the first 8 bytes of the file as a raw `u64` with no bounds check.
- [searchlite-core/src/index/terms.rs:42](searchlite-core/src/index/terms.rs#L42) — `let mut pairs = Vec::with_capacity(term_count as usize);` runs **before** the per-entry loop that would otherwise surface a short file.
- [searchlite-core/src/index/terms.rs:10-23](searchlite-core/src/index/terms.rs#L10-L23) — `write_terms` writes the 8-byte `term_count` header **before** computing `crc = checksum(&buf)`, so the CRC covers only the payload (`[varint_len][term_bytes][u64 offset]` repeated). The header is intentionally outside the inner integrity check.
- Reached from [searchlite-core/src/index/segment.rs:1296](searchlite-core/src/index/segment.rs#L1296) (`SegmentReader::open`) on every segment open — i.e. every reader/writer creation against the affected segment.

Compare against the fix in BUG-012 / #150 (`read_fields` in [searchlite-core/src/index/fastfields.rs](searchlite-core/src/index/fastfields.rs)), which introduced a `checked_count(count, min_stride, remaining)` helper that validates the count against the bytes still available in the buffer **before** passing it to `Vec::with_capacity`. That helper is not used in `terms.rs`.

## Reproduction / evidence

`read_terms` walks the file in this order ([searchlite-core/src/index/terms.rs:26-77](searchlite-core/src/index/terms.rs#L26-L77)):

```
u64    term_count       <-- untrusted, NOT covered by inner CRC
... payload ...         <-- covered by inner CRC
u32    crc              <-- inner CRC, over payload only
```

The inner CRC check at [searchlite-core/src/index/terms.rs:35-40](searchlite-core/src/index/terms.rs#L35-L40):

```rust
let payload = &buf[8..];
let (data, crc_bytes) = payload.split_at(payload.len() - 4);
let expected = u32::from_le_bytes([crc_bytes[0], crc_bytes[1], crc_bytes[2], crc_bytes[3]]);
let actual = checksum(data);
if expected != actual {
  bail!("terms file at {path:?} failed checksum validation");
}
```

`data` is `&buf[8 .. buf.len() - 4]` — the 8-byte `term_count` header is *not* part of `data`, and the matching write-side computation at [searchlite-core/src/index/terms.rs:19](searchlite-core/src/index/terms.rs#L19) confirms it: `let crc = checksum(&buf);` where `buf` was assembled starting from the first per-term entry, never including the header.

The Vec::with_capacity at line 42 then runs unconditionally:

```rust
let mut pairs = Vec::with_capacity(term_count as usize);
for _ in 0..term_count {
  let (len, consumed) = read_u64(&data[cursor..])?;
  ...
}
```

`(String, u64)` is 32 bytes on 64-bit (`String` = `{ ptr, len, cap } = 24 B`, plus `u64` = 8 B, aligned to 8 B). Crafting an inner-CRC-valid file with `term_count = u64::MAX` therefore asks the global allocator for `32 × u64::MAX` bytes, which:

- On stable Rust hits `RawVec::with_capacity_in`'s capacity-overflow check and panics with `capacity overflow` — process aborts.
- For more "moderate" tampered values (e.g. `term_count = 4_000_000_000`) the request becomes ~128 GiB; the allocator either returns a hard error and `handle_alloc_error` aborts, or "succeeds" under aggressive overcommit and the process is later OOM-killed when the pages are touched.

A failing-test sketch that a fix should make pass (modeled on the regression test for BUG-205):

```rust
#[test]
fn read_terms_rejects_oversized_term_count_on_short_file() {
    let dir = tempdir().unwrap();
    let path = dir.path().join("terms");
    let storage = crate::storage::FsStorage::new(dir.path().to_path_buf());

    // Write a valid one-term file, then corrupt only the 8-byte term_count
    // header and recompute the *outer* CRC scope (which deliberately excludes
    // the header) so the existing CRC check still passes.
    write_terms(&storage, &path, &[("only".to_string(), 1)]).unwrap();
    let mut raw = std::fs::read(&path).unwrap();
    // Overwrite the 8-byte header with u64::MAX.
    raw[..8].copy_from_slice(&u64::MAX.to_le_bytes());
    std::fs::write(&path, raw).unwrap();

    // The inner CRC still matches because it never covered the header.
    // The read must error out without committing a 128+ GiB allocation.
    let err = read_terms(&storage, &path)
        .expect_err("oversized term_count must be rejected");
    let msg = format!("{err:#}");
    assert!(
        msg.to_lowercase().contains("term_count") || msg.to_lowercase().contains("count"),
        "unexpected error: {msg}"
    );
}
```

## Threat model and reachability

There are two layers of integrity checking on the terms file:

1. **Inner CRC** ([terms.rs:35-40](searchlite-core/src/index/terms.rs#L35-L40)) — covers the payload but **not** the `term_count` header. Trivially defeated by editing only the header.
2. **Outer file CRC** computed by `verify_checksums` at [segment.rs:1214-1219](searchlite-core/src/index/segment.rs#L1214-L1219) over the entire on-disk file (`storage.read_to_end(path)`). This *does* cover the header, but only when `meta.checksums.get("terms")` returns `Some(..)`; the lookup at [segment.rs:1190](searchlite-core/src/index/segment.rs#L1190) silently skips verification when the entry is missing (legacy / hand-edited / merge-produced manifests with empty checksum maps — see e.g. [merge.rs:147](searchlite-core/src/index/merge.rs#L147) `checksums: HashMap::new()`).

In the case where `meta.checksums["terms"]` is present, a 1-byte flip in the header changes the whole-file CRC32, so most random disk corruption is caught — same level of mitigation that #205 has against the postings file. The read path itself, however, still allocates first and validates the count against the byte budget never. That makes this a defense-in-depth gap of the same shape and severity as #205, in the same module family.

## Expected behavior

`read_terms` validates `term_count` against the bytes still available in the file (analogous to `checked_count` in `fastfields::read_fields`) before passing it to `Vec::with_capacity`, and returns an `Err` when the count cannot possibly be satisfied by the remaining payload. The minimum per-term stride is 1 byte for the varint length + 0 bytes for an empty term + 8 bytes for the `u64` offset = **9 bytes**, which is enough to reject the `u64::MAX` case without rejecting legitimate compact segments.

## Actual behavior

`Vec::with_capacity(term_count as usize)` commits a capacity proportional to the unvalidated 8-byte header before the per-entry loop runs. On a memory-constrained host the global allocator calls `handle_alloc_error` and aborts the process; on an overcommitted host the process is later killed by the OOM killer when the pages are touched. Either way, opening the affected index becomes impossible — a DoS on every reader and writer.

## Root cause (hypothesis)

When BUG-012 was addressed for `fastfields.rs`, the same idiom in `terms.rs` was not audited — likely because `terms.rs` is much smaller (170 lines) and visually appears to "have a CRC", which masks that the CRC is scoped only to the payload and not the `term_count` header. The byte-slice-based `checked_count` helper used by `fastfields` would plug in cleanly here, since `read_terms` already operates on a fully-buffered `Vec<u8>`.

## Suggested fix (optional)

The minimum per-term stride is the varint length (≥ 1 byte) + 8 bytes for the offset = 9 bytes. A `checked_count(term_count as usize, 9, data.len())` call right after the CRC check, before `Vec::with_capacity`, is enough to bound the allocation to a count the file could actually carry. Returning the declared count and the byte-budget ceiling in the error message would let operators correlate corruption against a segment id.

## Environment

- Commit: `5261c9f` (HEAD at time of hunt)
- Rust MSRV in repo: 1.88.0
- Element size verified: `size_of::<(String, u64)>() == 32` on 64-bit (3 × `usize` for `String` + 8 B for `u64`, no padding)

davidkelley:
## Summary

Fixes #207. `read_terms` previously sized `Vec::with_capacity` directly from the 8-byte `term_count` header with no validation against the bytes remaining in the file.

The inner CRC at `terms.rs:38-40` is intentionally scoped to the **payload only** (see the corresponding write-side `checksum(&buf)` call where `buf` starts at the first per-term entry), so a single-byte flip in the header — or a merge-produced segment whose outer manifest checksum map is empty (`merge.rs:147`) — would otherwise drive a multi-gigabyte allocation before the per-entry read loop ever discovered the file is too short. On a memory-constrained host `handle_alloc_error` aborts the process; on an overcommitted host it dies later to OOM-kill. Either way, a DoS on every reader/writer that opens the affected segment.

This mirrors the BUG-012 / #150 fix in `fastfields::read_fields` and the BUG-205 / #206 fix in `postings::read_at`, adapted for the buffered byte-slice interface used by `terms.rs`:

- A local `checked_count(count, min_stride, remaining)` helper rejects counts whose minimum per-term byte cost would exceed the remaining payload budget, before the count reaches `Vec::with_capacity`.
- Minimum per-term stride: **9 bytes** — 1-byte varint length (smallest LEB128 encoding) + 0 bytes for an empty term + 8 bytes for the `u64` offset.
- `data.len()` is exactly the byte budget the per-entry loop has left (payload excludes both the 8-byte header and the trailing 4-byte CRC), so the check is tight without rejecting legitimate compact segments.

## Test plan

- [x] `cargo test -p searchlite-core --lib index::terms::` — 5 tests pass (2 new + 3 pre-existing)
- [x] `cargo test --workspace --lib` — all tests pass (266 + 47 + 7 across crates, 0 failures)
- [x] New regression tests:
  - `read_terms_rejects_oversized_term_count_on_short_file` — `u64::MAX` header, exercises the `checked_mul` overflow branch
  - `read_terms_rejects_moderately_oversized_term_count` — `4_000_000_000` header, exercises the `needed > remaining` branch

### DIFF
--- a/searchlite-core/src/index/terms.rs
+++ b/searchlite-core/src/index/terms.rs
@@ -74,7 +74,18 @@ pub fn read_terms(storage: &dyn Storage, path: &Path) -> Result<TinyFst> {
   for _ in 0..term_count {
     let (len, consumed) = read_u64(&data[cursor..])?;
     cursor += consumed;
-    let end = cursor + len as usize;
+    // `len` is read from an untrusted varint on disk. On 32-bit targets
+    // `len as usize` would truncate; on any target `cursor + len as usize`
+    // could wrap past `usize::MAX` for a crafted `u64::MAX`-ish length,
+    // letting the subsequent bounds check pass on the wrapped value and
+    // panicking the slice index below. Use `try_from` + `checked_add` so
+    // the overflow surfaces as a structured error instead.
+    let len_usize = usize::try_from(len).map_err(|_| {
+      anyhow!("terms file at {path:?} declares term length {len} that exceeds usize")
+    })?;
+    let end = cursor.checked_add(len_usize).ok_or_else(|| {
+      anyhow!("terms file at {path:?} declares term length {len} that overflows cursor")
+    })?;
     if end > data.len() {
       bail!("terms file at {path:?} ended unexpectedly while reading term");
     }
@@ -89,7 +100,13 @@ pub fn read_terms(storage: &dyn Storage, path: &Path) -> Result<TinyFst> {
       })?
       .to_string();
     cursor = end;
-    if cursor + 8 > data.len() {
+    // Same wrap concern as above, but with a fixed 8-byte stride — still
+    // worth guarding defensively so the bounds check cannot be bypassed
+    // by a `cursor` near `usize::MAX`.
+    let offset_end = cursor
+      .checked_add(8)
+      .ok_or_else(|| anyhow!("terms file at {path:?} cursor overflow reading offset"))?;
+    if offset_end > data.len() {
       bail!("terms file at {path:?} ended unexpectedly while reading offset");
     }
     let offset = u64::from_le_bytes([
@@ -255,6 +272,44 @@ mod tests {
     assert!(
       msg.contains("term_count") && msg.contains("remain"),
       "expected bounds-check error, got: {msg}"
+    );
+  }
+
+  /// Defense-in-depth companion to BUG-207: a crafted per-term varint
+  /// length near `u64::MAX` must be rejected via a structured error
+  /// rather than wrapping `cursor + len as usize` and panicking on the
+  /// subsequent slice index. Addresses the panic vector flagged by
+  /// Copilot review on the BUG-207 PR.
+  #[test]
+  fn read_terms_rejects_oversized_per_term_length() {
+    let dir = tempdir().unwrap();
+    let path = dir.path().join("terms");
+    let storage = crate::storage::FsStorage::new(dir.path().to_path_buf());
+
+    // Assemble a valid terms file by hand whose single per-term varint
+    // encodes `u64::MAX` as the term byte length. The CRC is computed
+    // over the payload so the checksum guard passes and execution
+    // reaches the cursor arithmetic we want to exercise.
+    let mut payload = Vec::new();
+    write_u64(u64::MAX, &mut payload);
+    // No term bytes — the cursor-wrap guard fires before any term data
+    // could be read. Append a plausible 8-byte offset so the payload
+    // is not rejected for being structurally incomplete earlier on.
+    payload.extend_from_slice(&0u64.to_le_bytes());
+    let crc = checksum(&payload);
+
+    let mut raw = Vec::new();
+    // term_count = 1 — passes checked_count (1 * 9 B <= payload bytes).
+    raw.extend_from_slice(&1u64.to_le_bytes());
+    raw.extend_from_slice(&payload);
+    raw.extend_from_slice(&crc.to_le_bytes());
+    std::fs::write(&path, raw).unwrap();
+
+    let err = read_terms(&storage, &path).expect_err("oversized per-term length must be rejected");
+    let msg = format!("{err:#}").to_lowercase();
+    assert!(
+      msg.contains("term length"),
+      "expected per-term length error, got: {msg}"
     );
   }
 }
--- a/searchlite-core/src/index/terms.rs
+++ b/searchlite-core/src/index/terms.rs
@@ -1,11 +1,36 @@
 use std::path::Path;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 
 use crate::storage::Storage;
 use crate::util::checksum::checksum;
 use crate::util::fst::TinyFst;
 use crate::util::varint::{read_u64, write_u64};
+
+/// Reject `term_count` values that cannot possibly be backed by the bytes
+/// still available in the terms payload. `term_count` is an 8-byte header
+/// read directly from the file and is **not** covered by the inner CRC
+/// (the CRC is scoped to the payload only), so a single-byte flip in that
+/// header — or a legacy/merge-produced segment with an empty outer
+/// checksum map — would otherwise drive a multi-gigabyte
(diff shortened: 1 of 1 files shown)


davidkelley:
Thanks for the review. Addressed the flagged concern in c807adc:

- `let end = cursor + len as usize;` → `usize::try_from(len)` + `cursor.checked_add(len_usize)`, returning a structured `term length ... exceeds usize` / `overflows cursor` error instead of wrapping and panicking on the subsequent slice.
- Applied the same `checked_add(8)` pattern to the offset bounds check for symmetry.
- Added `read_terms_rejects_oversized_per_term_length` — crafts a CRC-valid file whose per-term varint encodes `u64::MAX` and asserts the decoder returns a structured error.

@codex review

PR #214, opened 2026-04-15
test(postings): add test for position delta overflow

### LINKED ISSUE #141: BUG-003: Unchecked u32 addition when reconstructing absolute positions from deltas
# BUG-003: Unchecked u32 addition when reconstructing absolute positions from deltas

**Category:** Reliability
**Severity:** Medium
**File(s):** searchlite-core/src/index/postings.rs
**Line(s):** 208-211

## Description
`PostingsReader::read_at` reconstructs absolute positions by accumulating per-entry deltas:
```rust
let mut acc = 0u32;
for _ in 0..count {
  acc += read_u32_var(file)?;
  positions.push(acc);
}
```
The `acc += ...` is an unchecked `u32` addition. A corrupted or adversarial segment file can provide deltas whose sum exceeds `u32::MAX`. In debug builds this panics; in release builds it silently wraps, producing `positions` that are monotonically decreasing and violating the invariant relied on by phrase matching in `query/phrase.rs` and by any highlighter that trusts positional order.

The symmetric encode path in `PostingsWriter::write_term` also uses unchecked subtraction (`let delta = pos - prev`) and is covered by BUG-004.

## Reproduction / Evidence
Feed a segment whose first two position deltas are each `u32::MAX / 2 + 1`. The second `acc += ...` wraps to 0 (or panics) — no error is surfaced to the caller, and downstream phrase matchers see `[2_147_483_648, 0]` as valid positions.

## Expected Behaviour
Accumulation must validate the sum fits in `u32`, returning an `Err` on overflow so the caller can reject the segment.

## Actual Behaviour
Silent wrap in release / panic in debug; invariants of downstream consumers are broken.

## Suggested Fix
```rust
let mut acc = 0u32;
for _ in 0..count {
  let delta = read_u32_var(file)?;
  acc = acc.checked_add(delta)
    .ok_or_else(|| anyhow::anyhow!("position delta sum overflows u32"))?;
  positions.push(acc);
}
```

davidkelley:
## Summary

This PR adds a regression test for the unchecked u32 addition overflow in position delta accumulation when decoding postings.

## Background

The `PostingsReader::read_at` function reconstructs absolute positions from delta values using an accumulator. Previously, this accumulation was done with unchecked addition (`acc += delta`), which could:

- **Debug builds**: Panic due to overflow
- **Release builds**: Silently wrap around, producing incorrect position values

The overflow protection (`checked_add`) was already implemented in the codebase, but there was no test specifically for this overflow case.

## Changes

- Added `read_at_rejects_position_delta_overflow` test that creates a postings segment with two deltas each equal to `u32::MAX / 2 + 1`
- The sum of these deltas would overflow u32 (`u32::MAX + 2`), and the test verifies that the reader returns an error with the message "position delta overflow while decoding postings"

## Test Plan

- Run `cargo test -p searchlite-core postings` - all 12 tests pass
- The new test specifically validates that overflow is detected and reported correctly

### DIFF
--- a/searchlite-core/src/index/postings.rs
+++ b/searchlite-core/src/index/postings.rs
@@ -678,4 +678,37 @@ mod tests {
       .collect();
     assert_eq!(collected, vec![(1, 2, 0), (7, 4, 0)]);
   }
+
+  /// Regression for the unchecked u32 addition overflow in position delta
+  /// accumulation. Two deltas each equal to u32::MAX / 2 + 1 would sum to
+  /// u32::MAX + 2, overflowing in release builds (producing 0) or panicking
+  /// in debug builds. The fix uses checked_add to return an error instead.
+  #[test]
+  fn read_at_rejects_position_delta_overflow() {
+    use crate::util::varint::write_u32_var;
+    use std::io::Cursor;
+    // Create a minimal valid postings segment header
+    let mut buf = Vec::new();
+    buf.extend_from_slice(&1u32.to_le_bytes()); // doc_freq = 1 (little-endian u32)
+    buf.push(1); // stored_positions = true
+    buf.extend_from_slice(&0u32.to_le_bytes()); // raw_block (no block meta)
+    buf.extend_from_slice(&0u32.to_le_bytes()); // max_doc_id
+    buf.extend_from_slice(&0f32.to_le_bytes()); // max_tf
+
+    // Add a single posting with two positions that overflow when accumulated
+    // Position 1: delta = u32::MAX / 2 + 1 = 2_147_483_648
+    // Position 2: delta = u32::MAX / 2 + 1 = 2_147_483_648
+    // Sum would be u32::MAX + 2 = 4_294_967_298, which overflows u32
+    let delta: u32 = u32::MAX / 2 + 1;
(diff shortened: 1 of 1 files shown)


davidkelley:
@Copilot Thank you for the detailed review! I've addressed all your comments:

1. **Arithmetic correction**: Updated the doc comment to correctly state that the sum is `2^32` (`u32::MAX + 1`), not `u32::MAX + 2`.

2. **Comment values**: Fixed the inline comments to match the correct sum value.

3. **term_freq consistency**: Changed `term_freq` from `1` to `2` to match the position count, making the test segment match what `PostingsWriter` would produce.

4. **Assertion precision**: Updated the assertion to check for the full error message substring `"position delta overflow while decoding postings"` instead of just `"overflow"`.

All changes are in commit `8f8014e`.

davidkelley (commented):
I've addressed all four Copilot comments in the latest commit (8f8014e):

1. **Arithmetic correction** - Fixed the doc comment to correctly state the sum is `2^32` (`u32::MAX + 1`)

2. **Comment values** - Updated inline comments to match the correct sum value

3. **term_freq consistency** - Changed `term_freq` from `1` to `2` to match the position count

4. **Assertion precision** - Updated the assertion to check for the full error message substring

Please resolve the comments on GitHub.

PR #235, opened 2026-04-15
fix(cli): stream add/delete input instead of loading whole file (BUG-017)

### LINKED ISSUE #155: BUG-017: CLI `add` and `delete` commands load the entire input file into memory before streaming
# BUG-017: CLI `add` and `delete` commands load the entire input file into memory before streaming

**Category:** Performance
**Severity:** Medium
**File(s):** searchlite-cli/src/main.rs
**Line(s):** 286-308, 310-333

## Description
`cmd_add` and `cmd_delete` both read their input files with `fs::read_to_string(...)`, materialising the entire NDJSON document file (or ID list) in RAM before iterating `.lines()`. For realistic bulk-load workloads — the common operator use case for the CLI — a 1 GiB docs file becomes a 1 GiB `String`, plus the JSON parse and the `Document` construction per line. A 10 GiB load (well within a single searchlite segment's capacity) aborts with a `bad alloc` before any document is queued.

The loop body already handles one line at a time, so the up-front load serves no purpose. A `BufReader::lines()` stream would keep resident memory proportional to the longest single JSON document rather than to the entire file.

## Reproduction / Evidence
```rust
// searchlite-cli/src/main.rs:286-308
fn cmd_add(index: &Path, doc_path: &Path, write_key: Option<&str>) -> Result<()> {
  let opts = options(index, false);
  let idx = Index::open(opts)?;
  let mut writer = idx.writer_with_key(write_key)?;
  let content =
    fs::read_to_string(doc_path).with_context(|| format!("reading docs from {doc_path:?}"))?;
  for (line_no, line) in content.lines().enumerate() {
    // ... per-line parse + add_document ...
  }
  ...
}

// searchlite-cli/src/main.rs:310-333 — identical pattern for cmd_delete
let content = fs::read_to_string(ids_path)
  .with_context(|| format!("reading document ids from {ids_path:?}"))?;
```

Reproducer: `yes '{"_id":"x","body":"hi"}' | head -c 4G > docs.ndjson; searchlite-cli add --index ... --docs docs.ndjson` — the process is killed by the OOM killer before the first document is ingested.

## Expected Behaviour
The CLI should stream input line-by-line using `BufReader`, keeping memory bounded regardless of total file size.

## Actual Behaviour
The entire file is resident in RAM before any work happens.

## Suggested Fix
```rust
use std::io::{BufRead, BufReader};

fn cmd_add(index: &Path, doc_path: &Path, write_key: Option<&str>) -> Result<()> {
  let opts = options(index, false);
  let idx = Index::open(opts)?;
  let mut writer = idx.writer_with_key(write_key)?;
  let file = fs::File::open(doc_path)
    .with_context(|| format!("reading docs from {doc_path:?}"))?;
  let reader = BufReader::new(file);
  for (line_no, line) in reader.lines().enumerate() {
    let line = line.with_context(|| format!("reading line {}", line_no + 1))?;
    if line.trim().is_empty() { continue; }
    let value: serde_json::Value = serde_json::from_str(&line)
      .with_context(|| format!("invalid JSON on line {}", line_no + 1))?;
    let mut fields = std::collections::BTreeMap::new();
    if let Some(obj) = value.as_object() {
      for (k, v) in obj { fields.insert(k.clone(), v.clone()); }
    }
    writer.add_document(&Document { fields })?;
  }
  println!("queued documents (upsert), run commit to persist");
  Ok(())
}
```
Apply the same transformation to `cmd_delete`.

davidkelley:
## Summary

Fixes BUG-017. `cmd_add` and `cmd_delete` in `searchlite-cli/src/main.rs` called `fs::read_to_string` before iterating `.lines()`, so the full NDJSON input file (or ID list) had to fit in RAM before a single document was queued. For a realistic operator bulk-load (the canonical use case of the CLI) a 1 GiB docs file became a 1 GiB `String` plus the per-line parse cost; a 10 GiB load — well inside a single segment's capacity — aborted with a bad alloc before the first record was ingested. The per-line loop only ever needs one line resident.

## What changed

- **`searchlite-cli/src/main.rs`** — replace `fs::read_to_string(path)` + `str::lines()` in both `cmd_add` and `cmd_delete` with `BufReader::new(File::open(path)).lines()`. Resident memory is now bounded by the longest single line rather than the total file size.
- **`searchlite-cli/src/main.rs`** — `str::lines()` collapses `\r\n` into a `\r`-free slice, but `BufReader::lines()` only splits on `\n`, so a trailing `\r` survives on CRLF inputs. Add an explicit `strip_suffix('\r')` on the `cmd_add` path (it was already present on `cmd_delete`) to keep cross-platform NDJSON handling identical to the pre-fix behaviour. Empty-line skip, 1-based line numbers in error context, and the `io::Error` surface for mid-file read failures are preserved.
- **`searchlite-cli/src/main.rs`** (tests) — add three regression tests:
  - `cmd_add_handles_crlf_line_endings_with_streaming_reader` — pins the CRLF handling and blank-line skip after the reader switch (32 CRLF-terminated docs + one blank CRLF line, asserts all 32 are searchable after commit).
  - `cmd_add_reports_line_number_on_invalid_json` — locks in that per-line error context still references the 1-based line number after the streaming refactor.
  - `cmd_delete_streaming_reports_line_number_on_invalid_id` — same guarantee on the `cmd_delete` path (invalid id on line 2 must surface as `invalid id on line 2: …`).

## Test plan

- [x] `cargo test -p searchlite-cli` — 7/7 pass (4 existing + 3 new).
- [x] `cargo clippy -p searchlite-cli --tests --no-deps -- -D warnings` clean.
- [x] `cargo build -p searchlite-cli` clean.

Closes #155

### DIFF
--- a/searchlite-cli/src/main.rs
+++ b/searchlite-cli/src/main.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeMap;
 use std::fs;
-use std::io::{self, Read};
+use std::io::{self, BufRead, BufReader, Read};
 use std::path::{Path, PathBuf};
 
 use anyhow::{bail, Context, Result};
@@ -287,9 +287,14 @@ fn cmd_add(index: &Path, doc_path: &Path, write_key: Option<&str>) -> Result<()>
   let opts = options(index, false);
   let idx = Index::open(opts)?;
   let mut writer = idx.writer_with_key(write_key)?;
-  let content =
-    fs::read_to_string(doc_path).with_context(|| format!("reading docs from {doc_path:?}"))?;
-  for (line_no, line) in content.lines().enumerate() {
+  // Stream the NDJSON input line-by-line so memory stays bounded by the
+  // longest single document rather than by the total file size.
+  let file = fs::File::open(doc_path).with_context(|| format!("reading docs from {doc_path:?}"))?;
+  let reader = BufReader::new(file);
+  for (line_no, line) in reader.lines().enumerate() {
+    let line =
+      line.with_context(|| format!("reading docs from {doc_path:?} at line {}", line_no + 1))?;
+    let line = line.strip_suffix('\r').unwrap_or(&line);
     if line.trim().is_empty() {
       continue;
     }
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

---

PR #236, opened 2026-04-15
fix(merge): shrink batch instead of abandoning overflowing tier (BUG-008)

### LINKED ISSUE #146: BUG-008: TieredMergePolicy abandons merge entirely when the first N segments exceed max size, rather than reducing the batch
# BUG-008: TieredMergePolicy abandons merge entirely when the first N segments exceed max size, rather than reducing the batch

**Category:** Performance
**Severity:** Medium
**File(s):** searchlite-core/src/index/merge.rs
**Line(s):** 103-119

## Description
When a tier overflows, `find_merges` takes the first `min(tier.len(), max_merge_at_once)` segments, sums their effective doc counts, and *only* proceeds if the total fits under `max_merged_segment_docs`. If the total overshoots, the function returns `Vec::new()` — it never tries a smaller batch from the same tier, and never considers later (larger) tiers.

This creates an easy-to-hit silent stall. Example: `max_merge_at_once = 10`, `max_merged_segment_docs = 5_000_000`, and a tier containing 11 segments of roughly 800k docs each. The first 10 sum to 8M, which exceeds the cap, so no merge is emitted — yet a merge of 6 of them (≈4.8M) would be perfectly valid and would reduce the tier back under threshold. The policy will keep returning "no merges" on every subsequent tick while the tier continues to bloat from ongoing writes.

The accompanying test `skips_segments_at_max_size` only covers the filter at line 51 (individual segments already at the cap), not this aggregate failure mode.

## Reproduction / Evidence
```rust
// merge.rs:103-119
for tier in tiers.iter() {
  if tier.len() > self.segments_per_tier {
    let take = tier.len().min(self.max_merge_at_once);
    let ids: Vec<String> = tier[..take].iter().map(|s| s.id.clone()).collect();
    let total_docs: u64 = tier[..take]
      .iter()
      .map(|s| effective_doc_count(s) as u64)
      .sum();
    if total_docs <= self.max_merged_segment_docs as u64 {
      return vec![ids];
    }
    // <-- falls through to the next tier on failure, never reduces `take`
  }
}
```

Minimal reproducer:
```rust
let policy = TieredMergePolicy {
  max_merge_at_once: 10,
  segments_per_tier: 10,
  floor_segment_docs: 100_000,
  max_merged_segment_docs: 5_000_000,
};
let segments: Vec<_> = (0..11)
  .map(|i| make_segment(&format!("s{i}"), 800_000))
  .collect();
assert!(policy.find_merges(&segments).is_empty()); // stall
```

## Expected Behaviour
When the greedy pick overshoots, shrink the batch until it fits (`take -= 1`) or choose the largest prefix of the sorted tier whose cumulative doc count ≤ `max_merged_segment_docs`. Only give up on that tier if even the two smallest segments together exceed the cap.

## Actual Behaviour
The policy stalls: no merge is emitted despite the tier being over `segments_per_tier`, and incoming writes pile up further.

## Suggested Fix
Shrink the batch in place:
```rust
for tier in tiers.iter() {
  if tier.len() <= self.segments_per_tier {
    continue;
  }
  let mut take = tier.len().min(self.max_merge_at_once);
  while take >= 2 {
    let total_docs: u64 = tier[..take]
      .iter()
      .map(|s| effective_doc_count(s) as u64)
      .sum();
    if total_docs <= self.max_merged_segment_docs as u64 {
      let ids: Vec<String> = tier[..take].iter().map(|s| s.id.clone()).collect();
      return vec![ids];
    }
    take -= 1;
  }
}
```
Add a regression test covering the "11 × 800k with 5M cap" case above.

davidkelley:
## Summary

Fixes BUG-008 (#146). `TieredMergePolicy::find_merges` greedily took the first `min(tier.len(), max_merge_at_once)` segments from an overflowing tier and abandoned the tier entirely if their cumulative doc count exceeded `max_merged_segment_docs`. The common configuration of `max_merge_at_once = 10`, `max_merged_segment_docs = 5_000_000` against a tier of 11 × 800k segments therefore emitted no merge at all — even though a batch of 6 (~4.8M) would have fit. The tier would continue to bloat as further writes arrived, a silent stall that kept `find_merges` returning no-op until operators noticed via a segment-count alarm rather than via a merge error.

Because the segments in each tier are sorted by effective doc count ascending, the cumulative total is monotonic in `take`, so shrinking `take` until the prefix fits is both correct (largest valid prefix) and cheap.

## What changed

- **`searchlite-core/src/index/merge.rs`** — replace the single `let take = tier.len().min(self.max_merge_at_once); if total_docs <= cap { return vec![ids]; }` with a `while take >= 2` shrink loop that tries successively smaller prefixes until one fits the cap, and only gives up on the tier if even the two smallest segments together exceed the cap. Also flip the outer check to `if tier.len() <= segments_per_tier { continue; }` so the intent reads linearly and the shrink logic lives at the same indent as the rest of the pass. A block comment records why the ordering (ascending sort → monotonic total → safe to shrink) makes this correct.
- **`searchlite-core/src/index/merge.rs`** (tests) — three regression tests:
  - `shrinks_batch_when_greedy_pick_overshoots_max_merged_docs` — the exact 11 × 800k / 5M-cap scenario from the bug report. Asserts a batch of 6 is returned (the largest prefix that fits) and that the batch has no duplicate IDs.
  - `shrinks_batch_to_include_deleted_docs_in_cap_check` — tombstoned docs must drive the shrink. 3 × (900 total, 500 deleted) segments under a 1000 cap: raw totals would overshoot at any `take ≥ 2`, but effective totals let `take = 2` (800 live) fit. Locks in that the cap comparison uses `effective_doc_count`, matching the rest of the policy.
  - `gives_up_tier_when_even_two_smallest_exceed_cap` — pathological tier (3 × 700k under a 1M cap) where any pair sums to 1.4M. Must not emit a degenerate single-segment "merge" and must not loop forever. Exercises the `take < 2` exit path explicitly.

Verified by temporarily reverting the fix: the first two new tests fail on the old code (no merge emitted, vs expected 6-segment batch and 2-segment batch), the third passes either way — and all three pass with the fix applied. The six pre-existing merge-policy tests continue to pass unchanged, so no existing behaviour shifts.

## Test plan

- [x] `cargo test -p searchlite-core --lib index::merge` — 10/10 pass (7 existing + 3 new).
- [x] `cargo build -p searchlite-core` clean.
- [x] Verified the new regression tests catch the original bug by temporarily reverting the fix — `shrinks_batch_when_greedy_pick_overshoots_max_merged_docs` and `shrinks_batch_to_include_deleted_docs_in_cap_check` both fail on the unfixed code and pass with the fix applied.

Closes #146

### DIFF
--- a/searchlite-core/src/index/merge.rs
+++ b/searchlite-core/src/index/merge.rs
@@ -102,17 +102,26 @@ impl TieredMergePolicy {
 
     // Find the first tier that has too many segments.
     for tier in tiers.iter() {
-      if tier.len() > self.segments_per_tier {
-        let take = tier.len().min(self.max_merge_at_once);
-        let ids: Vec<String> = tier[..take].iter().map(|s| s.id.clone()).collect();
-        // Check that the merged result would not exceed the max size.
+      if tier.len() <= self.segments_per_tier {
+        continue;
+      }
+      // Greedy pick of the smallest `take` segments may overshoot
+      // `max_merged_segment_docs`. Because `tier` is sorted by effective doc
+      // count ascending, the cumulative total grows monotonically with `take`,
+      // so shrinking the batch until the total fits yields the largest valid
+      // prefix. Only give up on the tier if even the two smallest segments
+      // together exceed the cap.
+      let mut take = tier.len().min(self.max_merge_at_once);
+      while take >= 2 {
         let total_docs: u64 = tier[..take]
           .iter()
           .map(|s| effective_doc_count(s) as u64)
           .sum();
         if total_docs <= self.max_merged_segment_docs as u64 {
+          let ids: Vec<String> = tier[..take].iter().map(|s| s.id.clone()).collect();
           return vec![ids];
         }
+        take -= 1;
       }
     }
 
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

---

PR #237, opened 2026-04-15
fix(ffi): classify write-key auth errors via typed downcast (BUG-020)

### LINKED ISSUE #158: BUG-020: FFI classifies auth failures by substring-matching the error's Display form — renaming an error message silently flips the returned code
# BUG-020: FFI classifies auth failures by substring-matching the error's Display form — renaming an error message silently flips the returned code

**Category:** Reliability
**Severity:** Medium
**File(s):** searchlite-ffi/src/lib.rs
**Line(s):** 270-279, 320-329, 368-376, 419-429, 455-462, 491-499

## Description
Every mutating FFI entrypoint distinguishes the "write key required / incorrect" error code (`-8`) from the generic "writer open failed" code (`-4` or `-3`) by calling `err.to_string()` and testing whether it contains the literal substring `"write key"`. The contract with C callers is therefore:

> "`-8` means auth failure, `-4` means other failure" — but only while the upstream error text happens to include the literal ASCII sequence `write key`.

Any of the following silently breaks every downstream caller that branches on these codes:

- Renaming the error in `searchlite-core` (`"write key required"` → `"write_key required"` or `"Write Key required"` — the latter would still pass on one call site that uses `to_lowercase()` but not the other sites that don't).
- Adding a colon or period (some sites do `msg.contains("write key")`, not `starts_with`; fine today, but still fragile).
- Wrapping the original error with a new `with_context("additional context without the words")` — the lookup miss reclassifies auth errors as generic and a C client's retry logic (which may reopen the handle on `-4` but re-prompt for credentials on `-8`) does the wrong thing.
- Different FFI call sites are inconsistent: some use `msg.to_lowercase().contains("write key")` (lines 273, 370, 457) while others use the case-sensitive `msg.contains("write key")` (lines 323, 423, 493). A future refactor that unifies one side won't fix the other.

## Reproduction / Evidence
```rust
// searchlite-ffi/src/lib.rs:270-279 (add_json, case-insensitive)
let mut writer = match h.index.writer() {
  Ok(w) => w,
  Err(err) => {
    let msg = err.to_string();
    return if msg.to_lowercase().contains("write key") {
      -8
    } else {
      -4
    };
  }
};

// searchlite-ffi/src/lib.rs:320-329 (add_json_with_write_key, case-sensitive!)
Err(err) => {
  let msg = err.to_string();
  return if write_key.is_some() || msg.contains("write key") {
    -8
  } else {
    -4
  };
}
```

If `searchlite-core` changes the upstream string from `"write key required for this index"` to `"write-key required for this index"` (adding a hyphen), the case-sensitive sites immediately misclassify. If it changes to `"Write Key required"`, the case-insensitive sites keep working but a brand-new call site using `msg.contains` will not.

## Expected Behaviour
Authentication failures should be represented by a typed error variant in `searchlite-core` (e.g., `IndexError::WriteKeyRequired`, `IndexError::WriteKeyMismatch`), and the FFI should pattern-match on the variant rather than sniffing the display string.

## Actual Behaviour
The FFI <-> core error-code contract is tied to ad-hoc substring matching that can be silently broken by a message edit in a different crate.

## Suggested Fix
1. In `searchlite-core`, introduce a dedicated error type (or `thiserror` enum variant) for write-key failures:
   ```rust
   #[derive(Debug, thiserror::Error)]
   pub enum IndexError {
     #[error("write key required for this index")]
     WriteKeyRequired,
     #[error("write key does not match")]
     WriteKeyMismatch,
     // ...
   }
   ```
2. Wrap existing `anyhow!` call sites to use this typed error first, then let anyhow carry it through.
3. In the FFI, use `err.downcast_ref::<IndexError>()` instead of `err.to_string().contains(...)`:
   ```rust
   Err(err) => {
     if err.downcast_ref::<IndexError>().map_or(false, |e| {
       matches!(e, IndexError::WriteKeyRequired | IndexError::WriteKeyMismatch)
     }) { -8 } else { -4 }
   }
   ```
4. While fixing the FFI, also drop the `.to_string_lossy()` used on user-supplied paths (line 117) and query strings (line 527) — silent UTF-8 mangling there has the same "works until it doesn't" smell. Return an explicit error instead.

davidkelley:
## Summary

Fixes BUG-020 (#158). Every mutating FFI entrypoint (`searchlite_add_json`, `searchlite_add_json_batch`, `searchlite_commit`, and their `_with_write_key` variants) used to distinguish the auth error code `-8` from the generic writer-open error code `-4` / `-3` by calling `err.to_string()` and testing whether the Display form contained the literal substring `"write key"`. Different call sites even disagreed on casing (`.to_lowercase().contains(...)` vs plain `.contains(...)`), so a single message edit in `searchlite-core` could silently flip the returned error code for C callers whose retry logic branches on `-8` vs `-4`.

Introduce a dedicated `WriteKeyError` enum in `searchlite-core` with variants covering every auth failure (`Required`, `Mismatch(&'static str)`, `MetadataTampered`, `Empty`, `FeatureDisabled`), replace every `anyhow!(...)` / `bail!(...)` write-key call site with a typed return, and classify auth failures in the FFI via `anyhow::Error::downcast_ref::<WriteKeyError>()`. The FFI Display text is no longer part of the `-8` vs `-4`/`-3` contract, so renaming any message (or wrapping it in `.context(...)`) cannot silently reclassify an auth failure again.

## What changed

- **`searchlite-core/src/api/errors.rs`** — add the `WriteKeyError` enum (derives `thiserror::Error` + `Debug`) plus an `is_auth_variant()` predicate, and 4 unit tests that pin the downcast-through-context-chain property and stable Display strings.
- **`searchlite-core/src/api/mod.rs`** — re-export `WriteKeyError` alongside `AggregationError` and `PatchError` so FFI / downstream callers can name the type without reaching into `api::errors`.
- **`searchlite-core/src/api/writer.rs`** — `IndexWriter::new` now returns `WriteKeyError::Required` / `Mismatch(...)` / `MetadataTampered` / `FeatureDisabled` via typed `Err(...)` instead of `anyhow!` / `bail!`. `commit_with_merge` wraps its helpful "use `commit_with_merge_and_key` instead" hint as `.context(...)` on top of a typed `WriteKeyError::Required` so the message survives but classification is now reliable.
- **`searchlite-core/src/index/mod.rs`** — compaction and merge paths return the same typed variants (`Required`, `Mismatch("segment binding; index may be tampered")`, `FeatureDisabled`) in place of their `anyhow!("write key required for ...")` / `bail!(...)` predecessors.
- **`searchlite-core/src/util/write_key.rs`** — `derive_write_key_meta` returns `WriteKeyError::Empty`, `verify_write_key` returns `WriteKeyError::Mismatch("provided key did not match stored hash")` on a hash mismatch, and the `#[cfg(not(feature = "write-key"))]` stub returns `WriteKeyError::FeatureDisabled` — keeping all feature-gated paths on the same typed contract.
- **`searchlite-ffi/src/lib.rs`** — introduce a shared `classify_writer_err(&anyhow::Error, generic: c_int) -> c_int` helper that returns `-8` iff the error downcasts to `WriteKeyError` and `generic` otherwise, then call it from every mutating entrypoint. The explicit `write_key.is_some() → -8` bias on the `_with_write_key` variants is preserved as a strict pre-check so an intentional caller passing a key still gets `-8` on any writer-open failure (previous behaviour). 5 new FFI regression tests cover the helper directly (typed error recognised, context-wrapped error still recognised, unrelated error is NOT -8, a message that merely contains `"write key"` is NOT -8) plus an end-to-end test that exercises all three mutating entrypoints against a protected index without a key and asserts `-8` is returned.

## Test plan

- [x] `cargo test -p searchlite-core --lib api::errors` — 4/4 pass (new).
- [x] `cargo test -p searchlite-core` — full core test matrix green (lib + all integration suites, including the write-key integration flows).
- [x] `cargo test -p searchlite-ffi` — 19/19 pass (14 existing + 5 new regression tests).
- [x] `cargo test --workspace` — green across searchlite-core, searchlite-ffi, searchlite-cli, searchlite-http, searchlite-wasm, integration.
- [x] `cargo clippy -p searchlite-ffi --tests --no-deps -- -D warnings` clean.
- [x] Verified the new regression tests catch the original bug: the `downcast_recovers_variant_through_context_chain` and `classifies_typed_write_key_error_even_when_context_is_added` tests fail under the old substring-match approach when `.context(...)` is stacked on top of the error (which is exactly the failure mode BUG-020 flagged).

Closes #158

### DIFF
--- a/searchlite-core/src/api/errors.rs
+++ b/searchlite-core/src/api/errors.rs
@@ -24,3 +24,115 @@ pub enum PatchError {
   #[error("vector fields are not supported for updates")]
   VectorFieldsUnsupported,
 }
+
+/// Typed errors for the write-key auth path. Callers (notably the FFI) match
+/// on these variants via `anyhow::Error::downcast_ref` rather than sniffing
+/// the error's `Display` string — editing a message here will no longer
+/// silently reclassify an auth failure as a generic write failure downstream.
+#[derive(Debug, Error)]
+pub enum WriteKeyError {
+  /// The index enforces a write key, but the caller did not provide one.
+  #[error("write key required for this index")]
+  Required,
+
+  /// The index enforces a write key, but the caller provided the wrong one
+  /// (hash mismatch against the manifest) or a segment/WAL binding cannot be
+  /// re-derived from it (suggesting the index metadata was tampered).
+  #[error("write key does not match: {0}")]
+  Mismatch(&'static str),
+
+  /// The manifest records no write-key hash, yet at least one segment or
+  /// the WAL carries a binding — the manifest file was edited after the fact.
+  #[error("write key metadata missing but bindings exist; index metadata was likely tampered")]
+  MetadataTampered,
+
+  /// The caller supplied an empty string where a non-empty write key was
+  /// expected (rejected before any key material is written to disk).
+  #[error("write key cannot be empty")]
+  Empty,
+
+  /// The caller hit a write-key code path but the `write-key` Cargo feature
+  /// was not enabled at build time.
+  #[error("write-key feature is not enabled; rebuild with `--features write-key`")]
+  FeatureDisabled,
+}
+
+impl WriteKeyError {
+  /// Convenience predicate for the `Mismatch(_)` match arm so callers that
+  /// just want "is this an auth error?" don't have to name a specific reason.
+  pub const fn is_auth_variant(&self) -> bool {
+    matches!(
+      self,
+      WriteKeyError::Required
+        | WriteKeyError::Mismatch(_)
+        | WriteKeyError::MetadataTampered
+        | WriteKeyError::Empty
+        | WriteKeyError::FeatureDisabled
+    )
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use super::WriteKeyError;
+
+  #[test]
+  fn downcast_recovers_variant_after_anyhow_conversion() {
+    let err: anyhow::Error = WriteKeyError::Required.into();
+    let inner = err.downcast_ref::<WriteKeyError>().expect("must downcast");
+    assert!(matches!(inner, WriteKeyError::Required));
+  }
+
+  #[test]
+  fn downcast_recovers_variant_through_context_chain() {
+    // anyhow::Context wrapping must not hide the typed error from
+    // downcast_ref — this is the property BUG-020's substring match
+    // could not honour (context() shadows the Display string).
+    let err: anyhow::Error = WriteKeyError::MetadataTampered.into();
+    let err = err.context("opening writer").context("during compaction");
+    let inner = err.downcast_ref::<WriteKeyError>().expect("must downcast");
+    assert!(matches!(inner, WriteKeyError::MetadataTampered));
+  }
+
+  #[test]
+  fn display_messages_are_stable_for_human_readers() {
+    // Pin the message text so accidental edits become visible in review.
+    // Downstream classification does NOT depend on these strings after
+    // BUG-020's fix, but library users still read the messages.
+    assert_eq!(
+      WriteKeyError::Required.to_string(),
+      "write key required for this index"
+    );
+    assert_eq!(
+      WriteKeyError::Empty.to_string(),
+      "write key cannot be empty"
+    );
+    assert_eq!(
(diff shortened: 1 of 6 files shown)


davidkelley:
@codex review

---

PR #241, opened 2026-04-16
fix(schema): reject documents that omit non-nullable top-level fields

### LINKED ISSUE #224: fix(schema): validate_document silently accepts documents that omit non-nullable top-level fields
## Summary

`Schema::validate_document` never checks that all non-nullable top-level
fields are present. A document that omits a required (non-nullable) text,
keyword, or numeric field is accepted by validation, written by
`Writer::add_document`, and committed — even though `docs/schema.md`
states that *"every field defined in the schema is required -- Searchlite
rejects documents that omit a defined field."*

The same gap exists inside `collect_document` in the segment writer, so
the document ends up durably written without ever having the declared
required field indexed. Non-nullable **nested** sub-fields are checked
(see `NestedField::validate` at `searchlite-core/src/index/manifest.rs:1102`),
so this is a top-level-only omission.

## Affected code

- `searchlite-core/src/index/manifest.rs:349` — `Schema::validate_document`
  only iterates `doc.fields` (what the caller actually sent). There is no
  loop over `self.resolved_fields()` to assert required-and-missing.
- `searchlite-core/src/index/segment.rs:528` — `collect_document` mirrors
  the same "iterate present fields" shape: any non-nullable schema field
  that is absent is never visited, so no error is raised downstream either.
- `searchlite-core/src/api/writer.rs:259` and `:492` — both call sites rely
  on `validate_document` for this contract.
- Contrast with `searchlite-core/src/index/manifest.rs:1102` (and
  `searchlite-core/src/index/segment.rs:418`), which *do* enforce
  `missing required nested field …` for nested sub-properties.

## Reproduction / evidence

Against HEAD = `0967ca5`, with a minimal crate that depends on the
local `searchlite-core`:

```rust
use searchlite_core::api::types::{Document, IndexOptions, Schema, StorageType};
use searchlite_core::api::Index;

let schema_json = serde_json::json!({
    \"\$schema\": \"https://searchlite.dev/draft/2025/schema\",
    \"type\": \"object\",
    \"properties\": {
        \"body\": { \"type\": \"string\" }   // non-nullable by default
    }
});
let schema: Schema = serde_json::from_value(schema_json).unwrap();

// Resolved fields confirm body is parsed as non-nullable:
//   resolved_fields = [(\"body\", false)]

// Document that OMITS the required `body` field.
let doc = Document {
    fields: [(\"_id\".into(), serde_json::json!(\"doc-1\"))]
        .into_iter()
        .collect(),
};

// 1) Direct schema validation — per docs, should be Err(_). Observed: Ok(()).
assert!(schema.validate_document(&doc).is_err(), \"missing required field must error\");

// 2) Full writer pipeline — per docs, should reject the add. Observed:
//    Index::create -> writer.add_document(&doc) -> writer.commit() all succeed.
let tmp = tempfile::tempdir().unwrap();
let opts = IndexOptions {
    path: tmp.path().to_path_buf(),
    create_if_missing: true,
    enable_positions: true,
    bm25_k1: 1.2,
    bm25_b: 0.75,
    storage: StorageType::Filesystem,
};
let index = Index::create(tmp.path(), schema, opts).unwrap();
let mut writer = index.writer().unwrap();
writer.add_document(&doc).unwrap();  // BUG: should be Err
writer.commit().unwrap();            // BUG: should be Err
```

Actual program output:

```
resolved_fields = [(\"body\", false)]
BUG A: schema.validate_document accepted doc missing required \`body\`
BUG B: Index::add_document accepted doc missing required \`body\` (writer accepted)
BUG C: writer.commit succeeded even though required field was missing
```

Contrast: the analogous nested-field case is already covered and works
correctly — see the existing regression
`nested_fields_require_present_non_nullable_properties` in
`searchlite-core/src/index/manifest.rs` (around line 608+).

## Expected behavior

Per `docs/schema.md`:

> By default, every field defined in the schema is **required** --
> Searchlite rejects documents that omit a defined field.

`Schema::validate_document` (and by extension `Writer::add_document` /
`Writer::commit`) must return `Err(_)` when a document omits any
non-nullable field declared in the schema. The error message should
mirror the nested case for consistency, e.g. `missing required field
\`body\``.

## Actual behavior

Validation and the full write path silently accept the document. The
field is simply never indexed or stored for that doc, so subsequent
searches and filters on the declared-required field will behave as if
the field were optional — in direct contradiction of the documented
contract. Users who rely on the schema's "required" semantic for data
quality (e.g. to ensure every doc has a `title` or a `price`) get no
guarantee at all for top-level fields.

## Root cause (hypothesis)

`validate_document` walks `doc.fields` (what arrived) rather than
`self.resolved_fields()` (what is declared). A field that is declared
required but simply not present in the incoming `Document` never enters
the loop, so the required-ness check is never evaluated. The same
shape appears in `collect_document` in the segment writer. Both
predate — and were not updated by — the BUG-009 fix that hardened
nested-property validation in `#189`, which is why the nested path
correctly rejects missing required sub-fields but the top-level path
does not.

## Suggested fix (optional)

After the existing per-field loop in `validate_document`, add a second
pass over `self.resolved_fields()` (and top-level `self.nested_fields`)
that checks `doc.fields.contains_key(field.path)` for any field whose
`nullable` is `false`, and returns an error mirroring the nested
message. The corresponding guard inside `collect_document` then becomes
redundant, but adding it there as well keeps the segment writer honest
against direct callers that bypass the writer API.

## Environment

- Commit: `0967ca5`
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes #224. `Schema::validate_document` only iterated the fields present in the incoming document, so a document that omitted a declared non-nullable top-level text, keyword, numeric, or nested field was silently accepted by validation, written by `Writer::add_document`, and committed — in direct contradiction of the `docs/schema.md` contract that *"every field defined in the schema is required — Searchlite rejects documents that omit a defined field."*

Nested sub-fields were already enforced via `NestedField::validate` (see `searchlite-core/src/index/manifest.rs:1113`); the top-level path was not.

## What changed

- **`searchlite-core/src/index/manifest.rs`** — add a second pass in `validate_document` that walks declared `text_fields`, `keyword_fields`, `numeric_fields`, and `nested_fields`, bailing with `missing required field \`{name}\`` for any non-nullable field absent from the document. The error message mirrors the existing nested `missing required nested field …` wording for consistency. The doc-id field is skipped because it is already validated at the top of the function.
- **`searchlite-core/src/index/segment.rs`** — apply the same check in `collect_document` as defense-in-depth so direct segment-writer callers that bypass the schema validator still cannot write under-specified documents.
- **`searchlite-core/src/index/manifest.rs`** (tests) — add three regression tests:
  - `validate_document_rejects_missing_required_text_keyword_numeric` — document omitting all three field types must be rejected.
  - `validate_document_rejects_missing_required_nested_container` — document omitting a non-nullable nested container must be rejected.
  - `validate_document_allows_omitting_nullable_fields` — document omitting nullable text/keyword/numeric/nested fields must still be accepted.
- **`searchlite-core/tests/aggregations.rs`** — update three tests (`date_histogram_fixed_interval_respects_offset_and_missing`, `date_range_missing_and_keyed`, `extended_stats_and_value_count_include_missing`) to mark the sometimes-missing field as `nullable: true`. The aggregation `missing` default still applies to nullable fields, so assertions are unchanged.
- **`searchlite-core/tests/sorting.rs`** — update `sorts_numeric_and_missing_last` to mark the sometimes-missing `rating` field as `nullable: true`.
- **`searchlite-core/tests/partial_update.rs`** — update `update_set_unset_top_level_fields` to mark `body` as nullable, since unsetting a non-nullable field now correctly leaves an invalid document.
- **`searchlite-core/src/api/writer.rs`** — update `write_key_enforced_for_writer_open` to include the required `body` field in its test document.
- **`searchlite-core/src/index/mod.rs`** — update `compaction_requires_write_key_even_if_manifest_tampered` to include the required `body` field in its test document.
- **`examples/video-games/schema.json`** — mark 26 fields that are absent from some fixture documents as nullable, accurately reflecting the data shape.

## Test plan

- [x] `cargo test -p searchlite-core --lib index::manifest::tests` — 12/12 pass (9 existing + 3 new).
- [x] `cargo test -p searchlite-http`, `cargo test -p searchlite-cli`, `cargo test -p searchlite-ffi` — green.
- [x] `cargo test -p integration` — green (lifecycle matrix passes with updated fixture schema).
- [x] `cargo test --workspace` — green across all crates.

Closes #224

### DIFF
--- a/searchlite-core/src/index/fastfields.rs
+++ b/searchlite-core/src/index/fastfields.rs
@@ -1185,20 +1185,20 @@ fn write_presence(iter: impl Iterator<Item = bool>, buf: &mut Vec<u8>) {
 }
 
 fn doc_range(offsets: &[u32], doc: usize) -> Option<(usize, usize)> {
-  if offsets.len() < doc + 2 {
+  let start = *offsets.get(doc)? as usize;
+  let end = *offsets.get(doc.checked_add(1)?)? as usize;
+  if start > end {
     return None;
   }
-  let start = offsets[doc] as usize;
-  let end = offsets[doc + 1] as usize;
   Some((start, end))
 }
 
 fn object_range(offsets: &[u32], object_idx: usize) -> Option<(usize, usize)> {
-  if offsets.len() < object_idx + 2 {
+  let start = *offsets.get(object_idx)? as usize;
+  let end = *offsets.get(object_idx.checked_add(1)?)? as usize;
+  if start > end {
     return None;
   }
-  let start = offsets[object_idx] as usize;
-  let end = offsets[object_idx + 1] as usize;
   Some((start, end))
 }
 
@@ -1278,6 +1278,7 @@ fn read_fields(data: &[u8]) -> Result<HashMap<String, Column>> {
         for _ in 0..offsets_count {
           offsets.push(read_u32(&mut cursor, data)?);
         }
+        validate_monotonic_offsets(&offsets, "I64List")?;
         let total_vals = checked_count(
           *offsets.last().unwrap_or(&0) as usize,
           8,
@@ -1304,6 +1305,7 @@ fn read_fields(data: &[u8]) -> Result<HashMap<String, Column>> {
         for _ in 0..doc_offsets_count {
           doc_offsets.push(read_u32(&mut cursor, data)?);
         }
+        validate_monotonic_offsets(&doc_offsets, "I64Nested doc")?;
         let object_offsets_count = checked_count(
           checked_add_one(*doc_offsets.last().unwrap_or(&0) as usize)?,
           4,
@@ -1313,6 +1315,7 @@ fn read_fields(data: &[u8]) -> Result<HashMap<String, Column>> {
         for _ in 0..object_offsets_count {
           object_offsets.push(read_u32(&mut cursor, data)?);
         }
+        validate_monotonic_offsets(&object_offsets, "I64Nested object")?;
         let total_vals = checked_count(
           *object_offsets.last().unwrap_or(&0) as usize,
           8,
@@ -1359,6 +1362,7 @@ fn read_fields(data: &[u8]) -> Result<HashMap<String, Column>> {
         for _ in 0..offsets_count {
           offsets.push(read_u32(&mut cursor, data)?);
         }
+        validate_monotonic_offsets(&offsets, "F64List")?;
         let total_vals = checked_count(
           *offsets.last().unwrap_or(&0) as usize,
           8,
@@ -1385,6 +1389,7 @@ fn read_fields(data: &[u8]) -> Result<HashMap<String, Column>> {
         for _ in 0..doc_offsets_count {
           doc_offsets.push(read_u32(&mut cursor, data)?);
         }
+        validate_monotonic_offsets(&doc_offsets, "F64Nested doc")?;
         let object_offsets_count = checked_count(
           checked_add_one(*doc_offsets.last().unwrap_or(&0) as usize)?,
           4,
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

---

PR #256, opened 2026-04-16
fix(query): align normalize_pattern case-folding with default tokenizer (BUG-255)

### LINKED ISSUE #255: fix(query): normalize_pattern uses full Unicode lowering while default tokenizer uses ASCII-only, causing wildcard/regex queries to miss non-ASCII uppercase terms
## Summary

`Analyzer::normalize_pattern()` applies full Unicode case-folding via `str::to_lowercase()`, but the `default_tokenize()` function (used by the built-in default analyzer) applies ASCII-only case-folding via `char::to_ascii_lowercase()`. When a wildcard, regex, or multi-token pattern query contains non-ASCII uppercase characters (É, Ñ, Ü, Cyrillic, Greek, …), the pattern is lowered to a form that does not match the indexed terms, silently returning zero results for documents that should match.

This is the text-field analogue of the keyword-field divergence fixed in #212 / commit 8a85a67. That fix unified keyword case-folding via `fold_keyword()` but did not address the default tokenizer or its interaction with `normalize_pattern`.

## Affected code

- `searchlite-core/src/analysis/tokenizer.rs:13` — `default_tokenize()` uses `ch.to_ascii_lowercase()`, leaving non-ASCII uppercase code points unchanged (e.g. É stays É).
- `searchlite-core/src/analysis/analyzer.rs:42` — `normalize_pattern()` uses `pattern.to_lowercase()`, which folds all Unicode uppercase to lowercase (e.g. É → é).
- `searchlite-core/src/api/term_expansion.rs:274` — wildcard/regex patterns with metacharacters fall through to `normalize_pattern()`, producing a Unicode-lowered pattern that diverges from the ASCII-lowered indexed terms.
- `searchlite-core/src/api/term_expansion.rs:258` — zero-token fallback also uses `normalize_pattern()`.
- `searchlite-core/src/api/term_expansion.rs:280` — multi-token fallback also uses `normalize_pattern()`.

## Reproduction / evidence

Schema with a text field `title` using the default analyzer. Index one document:

```json
{"title": "RÉSUMÉ WRITING GUIDE"}
```

The default tokenizer indexes the terms: `["rÉsumÉ", "writing", "guide"]` — note `É` is preserved because `'É'.to_ascii_lowercase() == 'É'` (not ASCII).

**Wildcard query that should match but doesn't:**

```json
{"query": {"wildcard": {"title": {"value": "RÉS*"}}}}
```

1. `default_tokenize("RÉS*")` → token `"rÉs"` (single token, `*` stripped as separator).
2. `tokens.len() == 1` and `contains_wildcard_meta("RÉS*") == true`.
3. Falls through to `normalize_pattern("RÉS*")` → `"rés*"` (full Unicode: É → é).
4. Prefix expansion seeks terms starting with `"rés"`.
5. Indexed term is `"rÉsumÉ"` which starts with `"rÉs"` (É ≠ é at byte level: `0xC3 0x89` vs `0xC3 0xA9`).
6. **No match** — the document is silently dropped from results.

**Match query that does work (same document, same field):**

```json
{"query": {"match": {"title": "RÉSUMÉ"}}}
```

1. Both index and query use `default_tokenize`, producing `"rÉsumÉ"` on both sides.
2. **Match found** — the term lookup succeeds.

The two queries should agree on documents containing "RÉSUMÉ", but the wildcard query misses them.

## Expected behavior

A wildcard query `RÉS*` on a text field should find documents whose indexed tokens begin with the case-folded prefix `"rés"` (or equivalently, the tokenizer should produce `"résumé"` so that `"rÉs"` from `normalize_pattern` matches).

Either the tokenizer and the pattern normalizer must agree on the folding function, or the pattern normalizer must use the same folding as the tokenizer.

## Actual behavior

The indexed term is `"rÉsumÉ"` (ASCII-only folding preserves É) but the wildcard pattern is normalized to `"rés*"` (full Unicode folding lowers É → é). The byte sequences differ (`0xC3 0x89` vs `0xC3 0xA9`), so the prefix match fails and the document is silently omitted from results.

## Root cause (hypothesis)

`normalize_pattern()` was written to be a lightweight alternative to full re-tokenization for patterns that contain metacharacters. It checks whether the analyzer's tokenizer or filters would lowercase, and if so applies `str::to_lowercase()`. But the `Default` tokenizer lowercases via `to_ascii_lowercase()` — a weaker operation — while `normalize_pattern` unconditionally uses the stronger `to_lowercase()`. The two functions disagree on non-ASCII uppercase code points.

The `Unicode` tokenizer does not have this problem because it uses `word.to_lowercase()` (full Unicode) at tokenization time, matching `normalize_pattern`.

## Suggested fix

The most targeted fix: make `normalize_pattern()` match the tokenizer's actual folding behavior. For `TokenizerKind::Default`, apply `to_ascii_lowercase()` instead of `to_lowercase()`. For `TokenizerKind::Unicode` or when a `Lowercase` filter is present, continue using `to_lowercase()`.

Alternatively, upgrade `default_tokenize()` to use full Unicode lowering (mirroring the keyword-field fix from #212), which would fix the mismatch from the other direction and improve non-ASCII handling generally. This is the broader fix but changes indexed term forms, requiring a reindex.

## Environment

- Commit: 11f7be2
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- `normalize_pattern()` applied full Unicode lowering (`str::to_lowercase`) regardless of tokenizer, while the default tokenizer uses ASCII-only folding (`char::to_ascii_lowercase`). This caused wildcard/regex queries containing non-ASCII uppercase characters (É, Ñ, Ü, Cyrillic, Greek, …) to silently return zero results for documents that should match.
- Makes `normalize_pattern` match the tokenizer's actual folding strategy: ASCII-only for `Default`, full Unicode for `Unicode`/`Lowercase` filter, no folding for `Whitespace`.
- Adds unit tests covering all three tokenizer paths and an integration test reproducing the exact scenario from the bug report (wildcard `RÉS*` against indexed `RÉSUMÉ`).

## What changed

- **`searchlite-core/src/analysis/analyzer.rs`** — `normalize_pattern()` now branches on tokenizer kind: `Default` → `char::to_ascii_lowercase` per character; `Unicode` or `Lowercase` filter → `str::to_lowercase`; `Whitespace` (no filter) → identity.
- **`searchlite-core/src/api/term_expansion.rs`** — Two new unit tests pinning the wildcard and regex paths for non-ASCII uppercase with the default analyzer.
- **`searchlite-core/tests/regressions.rs`** — Integration test: indexes "RÉSUMÉ WRITING GUIDE", runs wildcard query `RÉS*`, asserts match. Also confirms query-string search agrees.

## Test plan

- [x] `cargo test -p searchlite-core --lib analysis::analyzer::tests::normalize_pattern` — 4/4 pass (new)
- [x] `cargo test -p searchlite-core --lib api::term_expansion::tests` — 39/39 pass (2 new + 37 existing)
- [x] `cargo test -p searchlite-core --test regressions wildcard_query_matches_non_ascii_uppercase` — 1/1 pass (new)
- [x] `cargo test -p searchlite-core` — full test suite green (347 unit + all integration)
- [x] `cargo fmt --all --check` clean
- [x] `cargo clippy -p searchlite-core --tests --no-deps -- -D warnings` clean

Closes #255

### DIFF
--- a/searchlite-core/src/analysis/analyzer.rs
+++ b/searchlite-core/src/analysis/analyzer.rs
@@ -30,16 +30,21 @@ impl Analyzer {
 
   /// Applies inexpensive, structure-preserving normalization suitable for patterns
   /// (e.g., wildcard/regex) without re-tokenizing or stripping delimiters.
+  ///
+  /// The folding strategy must match the tokenizer so that pattern terms align
+  /// with indexed terms. The `Unicode` tokenizer and the `Lowercase` filter
+  /// both apply full Unicode case-folding (`str::to_lowercase`), while the
+  /// `Default` tokenizer applies ASCII-only folding (`char::to_ascii_lowercase`).
   pub fn normalize_pattern(&self, pattern: &str) -> String {
-    let lowercases = matches!(
-      self.tokenizer,
-      TokenizerKind::Default | TokenizerKind::Unicode
-    ) || self
-      .filters
-      .iter()
-      .any(|f| matches!(f, TokenFilter::Lowercase));
-    if lowercases {
+    let has_unicode_lower = matches!(self.tokenizer, TokenizerKind::Unicode)
+      || self
+        .filters
+        .iter()
+        .any(|f| matches!(f, TokenFilter::Lowercase));
+    if has_unicode_lower {
       pattern.to_lowercase()
+    } else if matches!(self.tokenizer, TokenizerKind::Default) {
+      pattern.chars().map(|c| c.to_ascii_lowercase()).collect()
     } else {
       pattern.to_string()
     }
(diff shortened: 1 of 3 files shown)


davidkelley:
@codex review

PR #258, opened 2026-04-16
fix(aggs): preserve day-of-month in add_calendar for Month/Quarter/Year (BUG-257)

### LINKED ISSUE #257: fix(aggs): add_calendar forces day=1 in Month/Quarter/Year, producing misaligned fill-loop bucket keys when offset >= 24h
## Summary

`add_calendar()` unconditionally calls `.with_day(1)?` in its `Month`, `Quarter`, and `Year` calendar-interval variants. When a `date_histogram` aggregation specifies an `offset` of 24 hours or more (e.g. `"2d"`, `"36h"`), `bucket_start()` produces keys whose day-of-month is >1 (because the offset pushes the truncated-to-day-1 timestamp past midnight into the next day). The fill loop then feeds these offset-shifted keys back into `add_calendar`, which snaps the day back to 1 — producing bucket keys that no longer match the keys generated by `bucket_start()` for actual documents. The result is phantom empty buckets at the wrong timestamps and a malformed histogram.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3932` — `CalendarUnit::Month`: `date.with_year(year)?.with_month(month)?.with_day(1)?`
- `searchlite-core/src/query/aggs/mod.rs:3942` — `CalendarUnit::Quarter`: `date.with_year(year)?.with_month(month)?.with_day(1)?`
- `searchlite-core/src/query/aggs/mod.rs:3947` — `CalendarUnit::Year`: `.with_year(date.year() + 1)?.with_month(1)?.with_day(1)?`
- `searchlite-core/src/query/aggs/mod.rs:1543` — fill loop that chains `add_interval` calls

## Reproduction / evidence

Configure a `date_histogram` with `calendar_interval: "month"` and `offset: "2d"` (48 hours / 172,800,000 ms) on documents spanning several months with gaps between them.

**Trace of the bug:**

1. `bucket_start(june_ts, 2d, Calendar(Month))`:
   - `truncate_calendar(june_ts - 2d, Month)` → `2024-06-01T00:00:00Z`
   - `+ 2d` → **`2024-06-03T00:00:00Z`** (bucket key for June documents)

2. Fill loop calls `add_calendar(2024-06-03T00:00:00Z, Month)`:
   - Computes next month = 7 (July)
   - `date.with_year(2024)?.with_month(7)?` → `2024-07-03` (correct so far)
   - **`.with_day(1)?`** → `2024-07-01T00:00:00Z` (day snapped to 1)

3. Meanwhile, `bucket_start(july_ts, 2d, Calendar(Month))`:
   - `truncate_calendar(july_ts - 2d, Month)` → `2024-07-01T00:00:00Z`
   - `+ 2d` → **`2024-07-03T00:00:00Z`** (bucket key for July documents)

4. **Mismatch:** the fill loop inserts an empty bucket at `2024-07-01`, while documents create a populated bucket at `2024-07-03`. Both appear in the output — the histogram has two July entries (one phantom-empty at day 1, one real at day 3).

The issue compounds over subsequent months: the first bucket (from `bucket_start`) is correct at day 3, but every subsequent fill-loop bucket is misaligned at day 1.

**Minimal failing test sketch (not committed):**

```rust
#[test]
fn add_calendar_preserves_day_with_large_offset() {
    use chrono::{NaiveDate, Utc};
    // Input: 2024-06-03T00:00:00Z (day 3 from a 2-day offset)
    let dt = NaiveDate::from_ymd_opt(2024, 6, 3).unwrap()
        .and_hms_opt(0, 0, 0).unwrap();
    let ts = chrono::DateTime::<Utc>::from_naive_utc_and_offset(dt, Utc)
        .timestamp_millis();

    let next = add_calendar(ts, CalendarUnit::Month).unwrap();
    let expected = NaiveDate::from_ymd_opt(2024, 7, 3).unwrap()
        .and_hms_opt(0, 0, 0).unwrap();
    let expected_ts = chrono::DateTime::<Utc>::from_naive_utc_and_offset(expected, Utc)
        .timestamp_millis();

    // FAILS: actual is 2024-07-01, expected is 2024-07-03
    assert_eq!(next, expected_ts,
        "add_calendar(Month) must preserve the day-of-month");
}
```

## Expected behavior

`add_calendar` should advance the month/quarter/year while preserving the day-of-month (clamped to the target month's length if necessary). The bucket keys produced by the fill loop must match the keys produced by `bucket_start()` for the same period.

For the example above: `add_calendar(2024-06-03T00:00:00Z, Month)` should return `2024-07-03T00:00:00Z`.

## Actual behavior

`add_calendar` forces the day to 1 via `.with_day(1)?`, producing `2024-07-01T00:00:00Z`. This misaligns with `bucket_start`'s output of `2024-07-03T00:00:00Z`, creating phantom buckets in the histogram.

## Root cause (hypothesis)

The `.with_day(1)?` in `add_calendar` was carried over from `truncate_calendar` (where it is correct — truncation normalizes to the start of the period). But `add_calendar` should advance by one period while preserving the caller's day-of-month, not normalize it.

The issue was masked because:
1. The existing tests only use offsets < 24h (1 hour in the #251 regression tests), which keep the day at 1.
2. The `truncate_calendar` → `add_calendar` pipeline produces day=1 inputs when offset < 24h, so the `.with_day(1)?` is a no-op.

Note: the `.with_day(1)?` also serves a secondary purpose — it prevents `with_month()` from failing when the source day exceeds the target month's length (the same issue as BUG-233). The fix needs to normalize the day before `with_month()` to avoid that failure, but then restore the original day (clamped to the target month's max) afterward.

## Suggested fix (optional)

Replace the `.with_day(1)?` normalization with a day-preserving approach:

```rust
CalendarUnit::Month => {
    let mut month = date.month();
    let mut year = date.year();
    month += 1;
    if month > 12 { month = 1; year += 1; }
    let original_day = date.day();
    // Normalize to day 1 first to avoid with_month failure (BUG-233 pattern)
    let base = date.with_day(1)?.with_year(year)?.with_month(month)?;
    // Restore original day, clamped to the target month's length
    let max_day = days_in_month(year, month);
    base.with_day(original_day.min(max_day))?
}
```

Apply the same pattern to `Quarter` and `Year` variants. The `days_in_month` helper can use `chrono::NaiveDate::from_ymd_opt(year, month + 1, 1) - 1 day` or a similar idiom.

## Environment

- Commit: 61f149b
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- Fixes `add_calendar` forcing `day=1` in Month/Quarter/Year variants, which produced misaligned fill-loop bucket keys when `date_histogram` offset >= 24h
- Adds `last_day_of_month` helper to clamp day-of-month to the target month's length (avoiding both misalignment and the BUG-233 `with_month` failure on out-of-range days)
- Updates existing Year test expectation that relied on the old (buggy) month-snapping behavior

Closes #257

## Root cause

`add_calendar` unconditionally called `.with_day(1)?` in its Month, Quarter, and Year variants. When `bucket_start` produced keys with `day > 1` (due to offsets >= 24h like `"2d"` or `"36h"`), the fill loop generated phantom empty buckets at day 1 instead of matching the offset-shifted keys from `bucket_start`.

## Fix

Replace `.with_day(1)?` with a day-preserving approach:
1. Save the original day before changing year/month
2. Normalize to day 1 first (to prevent `with_month` from failing on out-of-range days — the BUG-233 pattern)
3. Restore the original day, clamped to the target month's max length via `last_day_of_month`

## Test plan

- [x] `add_calendar_preserves_day_with_large_offset` — verifies Month/Quarter/Year preserve day 3 from a 2-day offset
- [x] `add_calendar_fill_loop_aligned_with_multi_day_offset` — verifies chained add_calendar calls with 2-day offset match bucket_start round-trip
- [x] `add_calendar_clamps_day_to_target_month_length` — verifies Jan 31→Feb 28/29 clamping and Feb 29→Feb 28 across leap year boundary
- [x] All existing add_calendar tests continue to pass (sub-day time preservation, 1h-offset fill loop alignment)
- [x] Full `cargo test -p searchlite-core` (350 tests) passes
- [x] `cargo clippy -p searchlite-core -- -D warnings` clean

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3914,6 +3914,18 @@ fn truncate_calendar(value: i64, unit: CalendarUnit) -> Option<i64> {
   Some(chrono::DateTime::<Utc>::from_naive_utc_and_offset(start_dt, Utc).timestamp_millis())
 }
 
+fn last_day_of_month(year: i32, month: u32) -> u32 {
+  use chrono::Datelike;
+  if month == 12 {
+    31
+  } else {
+    chrono::NaiveDate::from_ymd_opt(year, month + 1, 1)
+      .and_then(|d| d.pred_opt())
+      .map(|d| d.day())
+      .unwrap_or(28)
+  }
+}
+
 fn add_calendar(value: i64, unit: CalendarUnit) -> Option<i64> {
   use chrono::{Datelike, Duration, Utc};
   let dt = chrono::DateTime::<Utc>::from_timestamp_millis(value)?;
@@ -3929,7 +3941,13 @@ fn add_calendar(value: i64, unit: CalendarUnit) -> Option<i64> {
         month = 1;
         year += 1;
       }
-      date.with_year(year)?.with_month(month)?.with_day(1)?
+      let original_day = date.day();
+      // Normalize to day 1 before changing year/month to prevent
+      // `with_month` from failing when the source day exceeds the
+      // target month's length (BUG-233 pattern).
+      let base = date.with_day(1)?.with_year(year)?.with_month(month)?;
+      let max_day = last_day_of_month(year, month);
+      base.with_day(original_day.min(max_day))?
     }
     CalendarUnit::Quarter => {
       let mut month = date.month();
@@ -3939,12 +3957,19 @@ fn add_calendar(value: i64, unit: CalendarUnit) -> Option<i64> {
         month -= 12;
         year += 1;
       }
-      date.with_year(year)?.with_month(month)?.with_day(1)?
+      let original_day = date.day();
+      let base = date.with_day(1)?.with_year(year)?.with_month(month)?;
+      let max_day = last_day_of_month(year, month);
+      base.with_day(original_day.min(max_day))?
     }
-    CalendarUnit::Year => date
-      .with_year(date.year() + 1)?
-      .with_month(1)?
-      .with_day(1)?,
+    CalendarUnit::Year => {
+      let new_year = date.year() + 1;
+      let original_day = date.day();
+      let original_month = date.month();
+      let base = date.with_day(1)?.with_year(new_year)?;
+      let max_day = last_day_of_month(new_year, original_month);
+      base.with_month(original_month)?.with_day(original_day.min(max_day))?
+    }
   };
   // Preserve the original time-of-day so that bucket keys remain aligned
   // with any sub-day offset applied by `bucket_start`. Previously this
@@ -4397,9 +4422,9 @@ mod tests {
       "add_calendar(Quarter) must preserve 01:00:00 offset"
     );
 
-    // Year: expect 2025-01-01T01:00:00Z
+    // Year: expect 2025-04-01T01:00:00Z (preserves month and day)
     let next_y = add_calendar(ts, CalendarUnit::Year).unwrap();
-    let expected_y = NaiveDate::from_ymd_opt(2025, 1, 1)
+    let expected_y = NaiveDate::from_ymd_opt(2025, 4, 1)
       .unwrap()
       .and_hms_opt(1, 0, 0)
       .unwrap();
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

davidkelley:
@codex review

PR #260, opened 2026-04-16
fix(analysis): skip edge_ngram emission for tokens shorter than min (BUG-259)

### LINKED ISSUE #259: fix(analysis): edge_ngram filter emits ngrams below configured min for short tokens
## Summary

The `edge_ngrams` function emits ngrams shorter than the user-configured `min` length when a token is shorter than `min`. Tokens shorter than `min` should produce no ngrams at all; instead, the function silently clamps `min` down to the token length and emits a ngram equal to the full token.

## Affected code

- `searchlite-core/src/analysis/analyzer.rs:430-431` — the `min` clamping logic in `edge_ngrams()`

## Reproduction / evidence

Configure an edge_ngram filter with `min=3, max=8` and analyze a token `"ab"` (2 characters):

```rust
let cfg = EdgeNgramConfig { min: 3, max: 8 };
// token: "ab" (len = 2)
let max = usize::min(8, 2);   // max = 2
let min = usize::min(3, 2);   // min = 2  <-- BUG: should be 3, clamped to 2
// guard: min == 0 || max == 0 → false, does NOT skip
// loop: size in 2..=2 → emits "ab" (length 2, violating cfg.min = 3)
```

A minimal failing test (not committed):

```rust
#[test]
fn edge_ngram_skips_tokens_shorter_than_min() {
    let tokens = analyze_with(
        vec![TokenFilterDef::EdgeNgram(EdgeNgramConfig { min: 3, max: 5 })],
        "ab",
    );
    // Token "ab" is 2 chars, shorter than min=3 → no ngrams should be emitted.
    assert!(tokens.is_empty()); // FAILS: tokens contains "ab"
}
```

## Expected behavior

When a token has fewer characters than `cfg.min`, the edge_ngram filter should produce no ngrams for that token (consistent with Elasticsearch's `EdgeNGramTokenFilter`, which drops tokens shorter than `min_gram`).

## Actual behavior

The token `"ab"` produces a single ngram `"ab"` (length 2), violating the user's configured minimum of 3. This causes:

1. Unexpected terms indexed below the minimum ngram length.
2. Unwanted search matches on very short prefixes that the user explicitly excluded via `min`.

For example, a user configuring `min=3` to avoid matching on 1–2 character prefixes would still see those matches because the short tokens get indexed.

## Root cause (hypothesis)

Line 431 computes `let min = usize::min(cfg.min, max)` where `max` was already clamped to the token length on line 430. When `len < cfg.min`, this silently reduces `min` to `len`. The zero-check guard on line 432 (`if min == 0 || max == 0`) only catches empty tokens (length 0), not tokens shorter than the configured minimum.

## Suggested fix

Add an early check before the clamping:

```rust
if len < cfg.min {
    continue;
}
```

This skips tokens shorter than the configured minimum, matching Elasticsearch semantics.

## Environment

- Commit: 22b989a
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- Fix `edge_ngrams()` to skip tokens shorter than the configured `min` length, matching Elasticsearch's `EdgeNGramTokenFilter` semantics
- Previously, `min` was silently clamped to the token length, causing short tokens (e.g. `"ab"` with `min=3`) to emit ngrams below the configured minimum
- Add tests for both the skip case (token < min) and the boundary case (token == min)

## Root cause

In `analyzer.rs:431`, the line `let min = usize::min(cfg.min, max)` reduces `min` to the token length when `len < cfg.min`. The zero-check guard only catches empty tokens, not tokens shorter than the configured minimum.

## Fix

Add an early-exit `if len < cfg.min { continue; }` before the clamping logic, so tokens shorter than the configured minimum produce no ngrams at all.

## Test plan

- [x] `cargo test -p searchlite-core -- edge_ngram` — 4 tests pass (2 new + 2 existing)
- [x] `cargo test -p searchlite-core` — full suite passes (349 tests)
- [x] `cargo clippy -p searchlite-core -- -D warnings` — clean

Closes #259

### DIFF
--- a/searchlite-core/src/analysis/analyzer.rs
+++ b/searchlite-core/src/analysis/analyzer.rs
@@ -585,7 +585,10 @@ mod tests {
       })],
       "ab",
     );
-    assert!(tokens.is_empty(), "tokens shorter than min should produce no ngrams");
+    assert!(
+      tokens.is_empty(),
+      "tokens shorter than min should produce no ngrams"
+    );
   }
 
   #[test]
--- a/searchlite-core/src/analysis/analyzer.rs
+++ b/searchlite-core/src/analysis/analyzer.rs
@@ -427,8 +427,11 @@ fn edge_ngrams(tokens: Vec<Token>, cfg: &EdgeNgramConfig) -> Vec<Token> {
   let mut out = Vec::new();
   for token in tokens.into_iter() {
     let len = token.text.chars().count();
+    if len < cfg.min {
+      continue;
+    }
     let max = usize::min(cfg.max, len);
-    let min = usize::min(cfg.min, max);
+    let min = cfg.min;
     if min == 0 || max == 0 {
       continue;
     }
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

davidkelley:
@codex review

PR #266, opened 2026-04-16
fix(aggs): bucket_script to_rpn silently accepts mismatched parentheses, producing wrong evaluation results (BUG-265)

### LINKED ISSUE #265: fix(aggs): bucket_script to_rpn silently accepts mismatched parentheses, producing wrong evaluation results
## Summary

The `to_rpn` shunting-yard implementation used by `eval_bucket_script` (pipeline aggregation expressions) silently accepts mismatched parentheses instead of returning `None`. An unmatched `)` prematurely drains the operator stack, changing operator precedence and producing a numerically wrong result. An unmatched `(` is silently discarded during cleanup, dropping the intended grouping. In both cases, the caller receives `Some(wrong_value)` rather than `None`, so no error is surfaced to the user.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3394` — `to_rpn`: the `RParen` branch does not check whether a matching `(` was found; the cleanup loop silently drops unmatched `(` without returning `None`.
- `searchlite-core/src/query/aggs/mod.rs:3312` — `eval_bucket_script`: calls `to_rpn`, which never returns `None`, so the only failure path is in `tokenize_script` or `eval_rpn`.

## Reproduction / evidence

Given variables `a=2, b=3, c=4`:

**Extra `)` changes precedence:**

Expression `"a + b) * c"` — the `)` at position 4 has no matching `(`. The `RParen` handler pops all operators (`+`) off the stack, producing RPN `[a, b, +, c, *]`. This evaluates as `(a + b) * c = 20`.

The intended expression (without the stray `)`) is `"a + b * c"`, which respects precedence and evaluates to `a + (b * c) = 14`.

Result: the stray `)` silently changes the answer from 14 to 20.

**Missing `)` drops grouping:**

Expression `"(a + b * c"` — the `(` has no matching `)`. During cleanup, the `(` is popped and discarded (`if op != '('`). The RPN becomes `[a, b, c, *, +]`, equivalent to `a + b * c = 14`.

If the user intended `"(a + b) * c"` and simply forgot the `)`, they get 14 instead of 20 with no error.

**Contrast with `script.rs`:**

The `shunting_yard` function in `searchlite-core/src/query/script.rs:338` (used by `script_score`) correctly detects both failure modes:
- Unmatched `)` → `bail!("mismatched parentheses in script_score")` (line 372)
- Unmatched `(` in cleanup → `bail!("mismatched parentheses in script_score")` (line 380)

## Expected behavior

`to_rpn` (or `eval_bucket_script`) should return `None` when the expression contains mismatched parentheses, so the pipeline aggregation reports a missing value for that bucket rather than silently producing a wrong number.

## Actual behavior

`to_rpn` always returns `Some(output)`. Mismatched parentheses silently alter operator precedence, producing a numerically incorrect result with no error indication.

## Root cause (hypothesis)

`to_rpn` was written as a simplified shunting-yard without parenthesis-balance validation. The `RParen` branch pops until `(` but does not check whether `(` was actually found — if not, the entire operator stack is drained, flattening all pending operators. The cleanup loop discards leftover `(` tokens instead of treating them as an error. The function signature returns `Option<Vec<ScriptToken>>` but the `None` path is unreachable — it always returns `Some`.

## Suggested fix

Mirror the validation from `shunting_yard` in `script.rs`:
1. In the `RParen` handler, track whether a matching `(` was found; return `None` if not.
2. In the cleanup loop, return `None` if a `(` is encountered (indicates an unmatched open paren).

## Environment

- Commit: 5403629
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- `to_rpn` shunting-yard implementation now returns `None` when parentheses are mismatched, instead of silently producing wrong results
- Unmatched `)` previously drained the operator stack and flattened precedence; now detected and rejected
- Unmatched `(` previously silently discarded during cleanup; now detected and rejected
- Mirrors the validated parenthesis-balance checks from `shunting_yard` in `script.rs`

## Changes

**`searchlite-core/src/query/aggs/mod.rs`**

- `RParen` handler: track whether a matching `(` was found; return `None` if not
- Cleanup loop: return `None` if a `(` remains on the operator stack (unmatched open paren)
- Added three tests: unmatched `)`, unmatched `(`, and matched parentheses (positive case)

## Test plan

- [x] `cargo test -p searchlite-core -- bucket_script` — 6 tests pass (3 existing + 3 new)
- [x] `cargo test -p searchlite-core` — full suite passes, no regressions
- [x] `cargo clippy -p searchlite-core --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean

Closes #265

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3413,19 +3413,25 @@ fn to_rpn(tokens: Vec<ScriptToken>) -> Option<Vec<ScriptToken>> {
       }
       ScriptToken::LParen => ops.push('('),
       ScriptToken::RParen => {
+        let mut found_lparen = false;
         while let Some(op) = ops.pop() {
           if op == '(' {
+            found_lparen = true;
             break;
           }
           output.push(ScriptToken::Op(op));
+        }
+        if !found_lparen {
+          return None;
         }
       }
     }
   }
   while let Some(op) = ops.pop() {
-    if op != '(' {
-      output.push(ScriptToken::Op(op));
+    if op == '(' {
+      return None;
     }
+    output.push(ScriptToken::Op(op));
   }
   Some(output)
 }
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

PR #268, opened 2026-04-16
fix(aggs): bucket_metric_value splits buckets_path on '.' breaking percentile key lookup for non-integer levels (BUG-267)

### LINKED ISSUE #267: fix(aggs): bucket_metric_value splits buckets_path on '.' breaking percentile key lookup for non-integer levels (e.g. 99.9)
## Summary

`bucket_metric_value` splits `buckets_path` on every `.` character, which corrupts the sub-aggregation key lookup for `percentiles` and `percentile_ranks` when the requested level contains a decimal point (e.g. `99.9`, `99.99`, `50.5`). All pipeline aggregations (`moving_avg`, `derivative`, `avg_bucket`, `sum_bucket`, `bucket_script`) and `bucket_sort` silently return `None`/null for affected references.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3534` — `path.split('.')` splits on every `.`, so `"my_pct.99.9"` becomes `["my_pct", "99", "9"]`
- `searchlite-core/src/query/aggs/mod.rs:3572` — `path.first().copied()` uses only `"99"` as the lookup key into the percentiles map
- `searchlite-core/src/query/aggs/mod.rs:3597` — `compute_percentiles_from_state` stores keys via `format!("{p}")`, which for `99.9_f64` produces `"99.9"`, not `"99"`
- `searchlite-core/src/query/aggs/mod.rs:3606` — `compute_percentile_ranks_from_state` has the same key format

All callers of `bucket_metric_value` are affected:
- `searchlite-core/src/query/aggs/mod.rs:3111` — `avg_bucket`
- `searchlite-core/src/query/aggs/mod.rs:3125` — `sum_bucket`
- `searchlite-core/src/query/aggs/mod.rs:3153` — `bucket_metric_series` (used by `derivative` and `moving_avg`)
- `searchlite-core/src/query/aggs/mod.rs:3269` — `bucket_script`
- `searchlite-core/src/query/aggs/mod.rs:3502` — `bucket_sort`

## Reproduction / evidence

Given a `date_histogram` with a `percentiles` sub-aggregation and a `moving_avg` pipeline referencing the 99.9th percentile:

```json
{
  "aggs": {
    "per_day": {
      "date_histogram": { "field": "timestamp", "calendar_interval": "day" },
      "aggs": {
        "latency_pct": {
          "percentiles": { "field": "latency_ms", "percents": [50, 95, 99.9] }
        },
        "smoothed_p999": {
          "moving_avg": { "buckets_path": "latency_pct.99.9", "window": 7 }
        }
      }
    }
  }
}
```

Trace through `bucket_metric_value("latency_pct.99.9")`:

1. **Line 3534:** `path.split('.')` → `["latency_pct", "99", "9"]`
2. **Line 3538:** `parts.remove(0)` → `agg_name = "latency_pct"`, remaining parts = `["99", "9"]`
3. **Line 3539:** `bucket.aggregations.get("latency_pct")` → finds the `Percentiles` response ✓
4. **Line 3540:** calls `extract_metric_from_response(agg, &["99", "9"])`
5. **Line 3572:** `path.first().copied()` → `"99"`
6. **Line 3573:** `vals.values.get("99")` → **`None`**, because the stored key is `"99.9"` (from `format!("{}", 99.9_f64)` at line 3597)

The `moving_avg` value for every bucket is `null`. The same failure occurs for any non-integer percentile level: `99.99` → lookup `"99"`, `50.5` → lookup `"50"`, etc.

Integer percentile levels are unaffected because `format!("{}", 95.0_f64)` produces `"95"` (no decimal point), so `split('.')` and `first()` yield the correct key.

## Expected behavior

`bucket_metric_value("latency_pct.99.9")` should resolve the sub-aggregation name as `"latency_pct"` and the metric key as `"99.9"`, successfully looking up the 99.9th percentile value from the percentiles response.

## Actual behavior

The `.` inside the percentile level `99.9` is treated as a path separator, splitting the key into `"99"` and `"9"`. The lookup for `"99"` fails silently, and every pipeline aggregation referencing a non-integer percentile returns `null`.

## Root cause (hypothesis)

`bucket_metric_value` uses an unqualified `path.split('.')` to separate the aggregation name from the metric sub-path. This works for simple cases like `"my_avg"` or `"my_stats.max"`, but breaks when the metric key itself contains a `.` — which is the case for non-integer percentile levels stored as their `Display` representation (e.g. `"99.9"`).

## Suggested fix

Use `splitn(2, '.')` instead of `split('.')` to split only on the **first** `.`, keeping the remainder intact as a single key:

```rust
fn bucket_metric_value(bucket: &BucketResponse, path: &str) -> Option<f64> {
    if path == "_count" {
        return Some(bucket.doc_count as f64);
    }
    let (agg_name, sub_path) = match path.split_once('.') {
        Some((name, rest)) => (name, Some(rest)),
        None => (path, None),
    };
    let agg = bucket.aggregations.get(agg_name)?;
    extract_metric_from_response(agg, sub_path)
}
```

Then update `extract_metric_from_response` to accept `Option<&str>` instead of `&[&str]`, using the full remainder string as the map key for `Percentiles`/`PercentileRanks`.

## Environment

- Commit: 5403629
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- `bucket_metric_value` used `path.split('.')` to separate the aggregation name from the metric sub-path, which breaks when the metric key itself contains a dot (e.g. non-integer percentile levels like `99.9`, `99.99`, `50.5`)
- Replace `split('.')` + `Vec` with `split_once('.')`, keeping the remainder intact as a single key string
- Update `extract_metric_from_response` to accept `Option<&str>` instead of `&[&str]`, using the full remainder as the map key for `Percentiles`/`PercentileRanks`

## Affected pipeline aggregations

All callers of `bucket_metric_value` are fixed: `moving_avg`, `derivative`, `avg_bucket`, `sum_bucket`, `bucket_script`, and `bucket_sort`.

## Test plan

- [x] Unit tests for `bucket_metric_value` with decimal percentile keys (`99.9`, `99.99`)
- [x] Unit test for `bucket_metric_value` with decimal percentile rank keys (`50.5`)
- [x] Unit tests confirming `Stats` sub-field lookup (`my_stats.max`) still works
- [x] Unit test for no-subpath default (`my_stats` → defaults to `avg`)
- [x] Integration test: `moving_avg` pipeline with `buckets_path: "latency_pct.99.9"` resolves correctly
- [x] Full `cargo test -p searchlite-core` passes
- [x] `cargo clippy` clean
- [x] `cargo fmt` clean

Closes #267

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3537,19 +3537,18 @@ fn bucket_metric_value(bucket: &BucketResponse, path: &str) -> Option<f64> {
   if path == "_count" {
     return Some(bucket.doc_count as f64);
   }
-  let mut parts: Vec<&str> = path.split('.').collect();
-  if parts.is_empty() {
-    return None;
-  }
-  let agg_name = parts.remove(0);
+  let (agg_name, sub_path) = match path.split_once('.') {
+    Some((name, rest)) => (name, Some(rest)),
+    None => (path, None),
+  };
   let agg = bucket.aggregations.get(agg_name)?;
-  extract_metric_from_response(agg, &parts)
+  extract_metric_from_response(agg, sub_path)
 }
 
-fn extract_metric_from_response(resp: &AggregationResponse, path: &[&str]) -> Option<f64> {
+fn extract_metric_from_response(resp: &AggregationResponse, path: Option<&str>) -> Option<f64> {
   match resp {
     AggregationResponse::Stats(stats) => {
-      let field = path.first().copied().unwrap_or("avg");
+      let field = path.unwrap_or("avg");
       match field {
         "avg" => Some(stats.avg),
         "min" => Some(stats.min),
@@ -3560,7 +3559,7 @@ fn extract_metric_from_response(resp: &AggregationResponse, path: &[&str]) -> Op
       }
     }
     AggregationResponse::ExtendedStats(stats) => {
-      let field = path.first().copied().unwrap_or("avg");
+      let field = path.unwrap_or("avg");
       match field {
         "avg" => Some(stats.avg),
         "min" => Some(stats.min),
@@ -3575,11 +3574,11 @@ fn extract_metric_from_response(resp: &AggregationResponse, path: &[&str]) -> Op
     AggregationResponse::ValueCount(val) => Some(val.value as f64),
     AggregationResponse::Cardinality(val) => Some(val.value as f64),
     AggregationResponse::Percentiles(vals) => {
-      let key = path.first().copied()?;
+      let key = path?;
       vals.values.get(key).copied()
     }
     AggregationResponse::PercentileRanks(vals) => {
-      let key = path.first().copied()?;
+      let key = path?;
       vals.values.get(key).copied()
     }
     AggregationResponse::AvgBucket(val) | AggregationResponse::SumBucket(val) => Some(val.value),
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

PR #272, opened 2026-04-16
fix(aggs): run bucket_sort after other pipeline aggregations (BUG-271)

### LINKED ISSUE #271: fix(aggs): bucket_sort runs before derivative/moving_avg/bucket_script, producing wrong sort order when sorting by pipeline-computed values
## Summary

`apply_pipeline_aggs` unconditionally runs `bucket_sort` **before** all other pipeline aggregations (derivative, moving_avg, bucket_script). When a `bucket_sort` sorts by a pipeline-computed field (e.g. the output of a derivative), the lookup finds no value in `bucket.aggregations` because the pipeline hasn't been applied yet. All buckets receive `Missing` for the sort field, falling back to a key-order comparison. Combined with a `size` truncation, this silently discards the buckets the user intended to keep.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3085-3148` — `apply_pipeline_aggs` function
- `searchlite-core/src/query/aggs/mod.rs:3090-3103` — first pass runs **only** `BucketSort`, sorting and truncating the bucket vec
- `searchlite-core/src/query/aggs/mod.rs:3105-3145` — second pass runs `AvgBucket`, `SumBucket`, `Derivative`, `MovingAvg`, `BucketScript` — **after** BucketSort has already sorted/truncated
- `searchlite-core/src/query/aggs/mod.rs:3504-3511` — `bucket_sort_value` returns `BucketSortComparable::Missing` when the pipeline field does not yet exist in `bucket.aggregations`

## Reproduction / evidence

Given a `date_histogram` with sub-aggregations:

```json
{
  "aggs": {
    "by_day": {
      "date_histogram": { "field": "ts", "calendar_interval": "day" },
      "aggs": {
        "daily_avg": { "avg": { "field": "latency" } },
        "rate_of_change": {
          "derivative": { "buckets_path": "daily_avg" }
        },
        "top_movers": {
          "bucket_sort": {
            "sort": [{ "rate_of_change": "desc" }],
            "size": 5
          }
        }
      }
    }
  }
}
```

Execution trace inside `apply_pipeline_aggs`:

1. **First pass** (lines 3090-3103): `bucket_sort_buckets` runs for `top_movers`.
   - For each bucket, `bucket_sort_value` looks up `"rate_of_change"` via `bucket_metric_value`.
   - `bucket.aggregations` contains only `daily_avg` (a non-pipeline agg); `rate_of_change` has not been inserted yet.
   - Every bucket returns `BucketSortComparable::Missing`.
   - The sort comparison treats all `Missing` values as `Equal`, falling back to the tiebreaker (`bucket_key_string` comparison → chronological key order).
   - `size: 5` truncates to the first 5 buckets by key.

2. **Second pass** (lines 3134-3135): `apply_derivative_pipeline` runs for `rate_of_change`.
   - Derivatives are computed only on the 5 surviving buckets.

The user asked for "the 5 days with the largest rate of change" but received "the 5 earliest days, with derivatives computed post-hoc."

## Expected behavior

`bucket_sort` should run **after** all other pipeline aggregations so it can access their computed values. The 5 buckets with the highest `rate_of_change` derivative values should be returned.

## Actual behavior

`bucket_sort` runs first when no derivative values exist, treats every bucket as `Missing` for the sort field, falls back to key-order sorting, and truncates. The result is 5 arbitrarily-selected (key-ordered) buckets with derivatives computed on only those 5 survivors.

## Root cause (hypothesis)

The `apply_pipeline_aggs` function intentionally runs `bucket_sort` first (lines 3090-3103) in a dedicated filter pass before all other pipelines. The likely intent was to let `bucket_sort` mutate the bucket vec (sort + truncate) before other pipelines iterate over it. However, this ordering prevents `bucket_sort` from accessing any pipeline-computed values, since they are only inserted during the second pass.

## Suggested fix

Move the `bucket_sort` execution to **after** all other pipeline aggregations:

```rust
fn apply_pipeline_aggs(...) -> BTreeMap<String, AggregationResponse> {
  let mut responses = BTreeMap::new();
  // 1. Run all non-BucketSort pipeline aggs first
  for (name, agg) in pipeline.iter() {
    match agg {
      Aggregation::AvgBucket(cfg) => { /* ... */ }
      Aggregation::SumBucket(cfg) => { /* ... */ }
      Aggregation::Derivative(cfg) => { /* ... */ }
      Aggregation::MovingAvg(cfg) => { /* ... */ }
      Aggregation::BucketScript(cfg) => { /* ... */ }
      Aggregation::BucketSort(_) => {} // deferred
      _ => {}
    }
  }
  // 2. Then run BucketSort, which can now see pipeline values
  for (name, agg) in pipeline.iter()
      .filter(|(_, a)| matches!(a, Aggregation::BucketSort(_)))
  {
    if let Aggregation::BucketSort(cfg) = agg {
      bucket_sort_buckets(buckets, cfg);
      responses.insert(name.clone(), AggregationResponse::BucketSort { ... });
    }
  }
  responses
}
```

Note: this reordering means `AvgBucket`/`SumBucket` will now reduce the full bucket list before truncation. If the current "reduce-after-truncation" behavior is desired, those reductions should also move after `bucket_sort`. The fix should match Elasticsearch semantics where `bucket_sort` runs after other sibling pipelines.

## Environment

- Commit: b8d57bd
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- Moves `bucket_sort` execution in `apply_pipeline_aggs` to run **after** all other pipeline aggregations (derivative, moving_avg, bucket_script, avg_bucket, sum_bucket), so it can access pipeline-computed values when sorting
- Previously, `bucket_sort` ran first in a dedicated pass, found no pipeline values, returned `Missing` for every sort field, and fell back to key-order sorting — silently producing wrong results when sorting by pipeline-computed fields like derivatives
- Adds a regression test that creates a histogram with a derivative pipeline and a `bucket_sort` sorting by derivative value with `size: 2`, verifying the correct buckets survive truncation

Closes #271

## Test plan

- [x] New `bucket_sort_runs_after_derivative_pipeline` regression test passes
- [x] All 42 aggregation tests pass
- [x] Full `cargo test -p searchlite-core` suite passes with 0 failures

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3087,21 +3087,6 @@ fn apply_pipeline_aggs(
   buckets: &mut Vec<BucketResponse>,
 ) -> BTreeMap<String, AggregationResponse> {
   let mut responses = BTreeMap::new();
-  for (name, agg) in pipeline
-    .iter()
-    .filter(|(_, a)| matches!(a, Aggregation::BucketSort(_)))
-  {
-    if let Aggregation::BucketSort(cfg) = agg {
-      bucket_sort_buckets(buckets, cfg);
-      responses.insert(
-        name.clone(),
-        AggregationResponse::BucketSort {
-          from: cfg.from.unwrap_or(0),
-          size: cfg.size,
-        },
-      );
-    }
-  }
   for (name, agg) in pipeline.iter() {
     match agg {
       Aggregation::AvgBucket(cfg) => {
@@ -3142,6 +3127,21 @@ fn apply_pipeline_aggs(
       }
       Aggregation::BucketSort(_) => {}
       _ => {}
+    }
+  }
+  for (name, agg) in pipeline
+    .iter()
+    .filter(|(_, a)| matches!(a, Aggregation::BucketSort(_)))
+  {
+    if let Aggregation::BucketSort(cfg) = agg {
+      bucket_sort_buckets(buckets, cfg);
+      responses.insert(
+        name.clone(),
+        AggregationResponse::BucketSort {
+          from: cfg.from.unwrap_or(0),
+          size: cfg.size,
+        },
+      );
     }
   }
   responses
--- a/searchlite-core/tests/aggregations.rs
+++ b/searchlite-core/tests/aggregations.rs
@@ -4339,3 +4339,172 @@ fn significant_terms_preserves_high_significance_low_frequency_terms() {
     panic!("expected significant_terms response");
   }
 }
+
+#[test]
+fn bucket_sort_runs_after_derivative_pipeline() {
+  let tmp = tempfile::tempdir().unwrap();
+  let path = tmp.path().to_path_buf();
+  let mut schema = Schema::default_text_body();
+  schema.numeric_fields.push(NumericField {
+    name: "val".into(),
+    i64: false,
+    fast: true,
+    stored: false,
+    nullable: false,
+  });
+  schema.numeric_fields.push(NumericField {
+    name: "metric".into(),
+    i64: false,
+    fast: true,
+    stored: false,
+    nullable: false,
+  });
+  let idx = Index::create(
+    &path,
+    schema,
+    IndexOptions {
+      path: path.clone(),
+      create_if_missing: true,
+      enable_positions: true,
+      bm25_k1: 0.9,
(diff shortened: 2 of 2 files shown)


davidkelley:
@codex review

---

PR #276, opened 2026-04-16
fix(aggs): handle unary negation of variables in bucket_script (BUG-275)

### LINKED ISSUE #275: fix(aggs): bucket_script tokenize_script silently misparses unary negation of variables, producing None instead of correct values
## Summary

`tokenize_script` in the bucket_script evaluator only recognizes unary minus when followed by a digit or decimal point. When unary minus precedes a variable name (e.g., `-a`, `a * -b`, `a - -b`), the minus is tokenized as a binary operator instead of a unary negation. This produces an operator-heavy token stream that fails during RPN evaluation, causing `eval_bucket_script` to silently return `None` for every bucket — yielding missing values where the user expects numeric results.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3433-3441` — the `looks_numeric` guard in `tokenize_script` requires the character after `-` to be a digit or `.`; a variable-name character (alphabetic/underscore) causes the guard to fail and the `-` falls through to the binary-operator branch at line 3479.
- `searchlite-core/src/query/aggs/mod.rs:3545-3575` — `eval_rpn` silently returns `None` when the malformed RPN token stream underflows the operand stack (too many operators, too few operands).
- `searchlite-core/src/query/aggs/mod.rs:3364-3399` — `apply_bucket_script_pipeline` propagates the `None` as a missing value for every bucket's BucketScript aggregation.

## Reproduction / evidence

**Input script:** `"a * -b"` with `vars = {a: 2.0, b: 3.0}`

**Tokenization trace:**
1. `a` → `Var("a")`, `expect_unary = false`
2. `*` → `Op('*')`, `expect_unary = true`
3. `-`: `expect_unary = true`, but next char is `b` (not a digit/`.`), so `looks_numeric = false`. Falls through to operator match → `Op('-')`, `expect_unary = true`
4. `b` → `Var("b")`

**Resulting tokens:** `[Var("a"), Op('*'), Op('-'), Var("b")]`

**Shunting-yard RPN conversion:**
- `Var("a")` → output
- `Op('*')` → ops stack
- `Op('-')`: precedence(`*`)=2 ≥ precedence(`-`)=1 → pop `*` to output, push `-`
- `Var("b")` → output
- Drain: pop `-` to output
- **RPN:** `[Var("a"), Op('*'), Var("b"), Op('-')]`

**RPN evaluation:**
1. Push `a=2.0` → stack: `[2.0]`
2. `Op('*')`: pop `b=2.0`, pop `a`… stack underflow → **returns `None`**

**Expected result:** `2.0 * (-3.0) = -6.0`

The same class of failure affects any expression with unary negation on a variable: `-a`, `-a + b`, `a - -b`, `a / -b`, `(-b)`, etc.

## Expected behavior

`eval_bucket_script("a * -b", {a: 2.0, b: 3.0})` should return `Some(-6.0)`.

Either:
1. `tokenize_script` should emit a synthetic `Number(0)` and `Op('-')` when it encounters unary minus before a variable (effectively rewriting `-x` as `0 - x`), OR
2. A dedicated `ScriptToken::Neg` unary operator should be supported in the tokenizer, RPN converter, and evaluator.

## Actual behavior

`eval_bucket_script("a * -b", {a: 2.0, b: 3.0})` returns `None`. All buckets get `BucketScript { value: None }` and any downstream pipeline aggregation referencing the BucketScript also sees `None`.

## Root cause (hypothesis)

The `looks_numeric` guard at line 3433-3441 checks whether the character after `-` is `is_ascii_digit() || '.'`. When the next character is alphabetic (a variable name), the check fails and the `-` is routed to the binary-operator arm at line 3479. The Shunting-yard algorithm then treats both `*` and `-` as binary operators, producing RPN with two consecutive binary ops that consume three operands from a stream that only provides two. The resulting stack underflow in `eval_rpn` is handled by returning `None` rather than panicking, so the failure is completely silent — no error is logged or surfaced to the caller.

## Suggested fix

The simplest correct fix is to handle unary minus in the tokenizer by emitting `Number(0.0)` followed by `Op('-')` when `expect_unary` is true and the `-` is NOT followed by a digit. This transforms `-x` into `0 - x` at the token level, which the existing Shunting-yard and RPN evaluator handle correctly. Alternatively, `(-1) * x` can be used, or a proper unary-negation token can be added.

A corresponding change to `to_rpn` or `eval_rpn` is not needed if the synthetic-zero approach is used.

## Environment

- Commit: d17df1b
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- Fixes `tokenize_script` in the bucket_script evaluator to correctly handle unary negation of variables (e.g., `-a`, `a * -b`, `a - -b`, `(-b)`, `a / -b`)
- Adds a dedicated `Neg` unary operator token with higher precedence than `*`/`/`, processed through the tokenizer, shunting-yard RPN converter, and RPN evaluator
- Previously, unary minus before a variable was misclassified as a binary operator, producing a malformed token stream that caused `eval_bucket_script` to silently return `None` for every bucket

## Root cause

The `looks_numeric` guard in `tokenize_script` only recognized unary minus when followed by a digit or `.`. When `-` preceded a variable name, it fell through to the binary operator branch, producing two consecutive binary operators in the token stream. The RPN evaluator's stack underflowed and silently returned `None`.

## Approach

Rather than the synthetic-zero rewrite (`-x` → `0 - x`) which breaks operator precedence for expressions like `a * -b`, this uses a proper unary negation operator (`Neg`) that:
1. Has higher precedence than all binary operators (precedence 3 vs 2 for `*`/`/`)
2. Is emitted by the tokenizer when `expect_unary` is true and `-` is not followed by a digit
3. Pops a single operand in `eval_rpn` and pushes its negation

## Test plan

- [x] `cargo test -p searchlite-core bucket_script` — 11 tests (6 existing + 5 new)
- [x] `cargo test -p searchlite-core` — full suite passes (367 tests)
- [x] `cargo clippy -p searchlite-core --all-targets -- -D warnings` — clean
- [x] `cargo fmt -p searchlite-core -- --check` — clean

New test cases cover: `a * -b`, `-a`, `a - -b`, `(-b)`, `a / -b`

Closes #275

https://claude.ai/code/session_01S13nj6nAPPTkfa1GufYLQ4

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3403,6 +3403,7 @@ enum ScriptToken {
   Number(f64),
   Var(String),
   Op(char),
+  Neg,
   LParen,
   RParen,
 }
@@ -3411,6 +3412,7 @@ fn op_precedence(op: char) -> u8 {
   match op {
     '+' | '-' => 1,
     '*' | '/' => 2,
+    '~' => 3,
     _ => 0,
   }
 }
@@ -3477,6 +3479,11 @@ fn tokenize_script(script: &str) -> Option<Vec<ScriptToken>> {
     }
     match ch {
       '+' | '-' | '*' | '/' => {
+        if ch == '-' && expect_unary {
+          tokens.push(ScriptToken::Neg);
+          chars.next();
+          continue;
+        }
         tokens.push(ScriptToken::Op(ch));
         chars.next();
         expect_unary = true;
@@ -3497,20 +3504,30 @@ fn tokenize_script(script: &str) -> Option<Vec<ScriptToken>> {
   Some(tokens)
 }
 
+fn pop_op(op: char) -> ScriptToken {
+  if op == '~' {
+    ScriptToken::Neg
+  } else {
+    ScriptToken::Op(op)
+  }
+}
+
 fn to_rpn(tokens: Vec<ScriptToken>) -> Option<Vec<ScriptToken>> {
   let mut output = Vec::new();
   let mut ops: Vec<char> = Vec::new();
   for token in tokens.into_iter() {
     match token {
       ScriptToken::Number(_) | ScriptToken::Var(_) => output.push(token),
+      ScriptToken::Neg => {
+        ops.push('~');
+      }
       ScriptToken::Op(op) => {
         while let Some(&top) = ops.last() {
           if top == '(' {
             break;
           }
           if op_precedence(top) >= op_precedence(op) {
-            output.push(ScriptToken::Op(top));
-            ops.pop();
+            output.push(pop_op(ops.pop().unwrap()));
           } else {
             break;
           }
@@ -3525,7 +3542,7 @@ fn to_rpn(tokens: Vec<ScriptToken>) -> Option<Vec<ScriptToken>> {
             found_lparen = true;
             break;
           }
-          output.push(ScriptToken::Op(op));
+          output.push(pop_op(op));
         }
         if !found_lparen {
           return None;
@@ -3537,7 +3554,7 @@ fn to_rpn(tokens: Vec<ScriptToken>) -> Option<Vec<ScriptToken>> {
     if op == '(' {
       return None;
     }
-    output.push(ScriptToken::Op(op));
+    output.push(pop_op(op));
   }
   Some(output)
 }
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

---

PR #286, opened 2026-04-17
fix(highlight): ensure fragment window fully contains match (BUG-285)

### LINKED ISSUE #285: fix(highlight): fragment window does not guarantee match containment when match length exceeds fragment_size/2
## Summary

`highlight_fragments` computes a fragment window of `fragment_size` bytes centered on `m.start()`, giving the match only `fragment_size / 2` bytes of forward coverage. When a match (typically a phrase query with many terms) is longer than `fragment_size / 2` bytes, its tail is truncated from the fragment. The subsequent `regex.replace_all` on the truncated fragment then fails to find the full match, so the fragment is returned **without any highlight tags** — a silent correctness failure.

## Affected code

- `searchlite-core/src/index/highlight.rs:50-51` — fragment window calculation

```rust
let raw_start = m.start().saturating_sub(opts.fragment_size / 2);
let raw_end = usize::min(text.len(), raw_start.saturating_add(opts.fragment_size));
```

The window spans `[m.start() - frag/2, m.start() + frag/2)`. The match spans `[m.start(), m.end())`. For the match to be fully contained, `m.end() - m.start() <= fragment_size / 2` must hold. When it does not, the tail of the match falls outside the fragment.

- `searchlite-core/src/index/highlight.rs:56-59` — regex replacement applied to the (possibly truncated) fragment

## Reproduction / evidence

Concrete failing scenario with `fragment_size = 100`:

```
text = "aaaaaaaaa bbbbbbbb cccccccc dddddddd eeeeeeee \
        the quick brown fox jumps over the lazy dog near the river yyyy"

phrase = ["the","quick","brown","fox","jumps","over","the","lazy","dog","near","the","river"]
```

The phrase match is ~57 bytes. It starts at byte ~46 in the text.

1. `m.start() = 46`, `m.end() = 103`, match length = 57 bytes.
2. `raw_start = 46 - 50 = 0` (saturating), `raw_end = min(len, 0 + 100) = 100`.
3. Fragment = `text[0..100]` — missing bytes 100..103 of the match.
4. Phrase regex `\bthe\W+quick\W+…\W+the\W+river\b` does **not** match inside the truncated fragment because "river" is only partially present.
5. `replace_all` returns the fragment unchanged — **no highlight tags**.

The bug triggers whenever `m.start() > fragment_size / 2` **and** `match_length > fragment_size / 2`. For the default `make_snippet` fragment size of 120, any phrase match longer than 60 bytes (~12 short words) that doesn't start near the beginning of the text will silently lose highlighting. Multi-byte UTF-8 text (accented Latin, CJK) reaches this threshold with fewer characters.

## Expected behavior

The fragment window should always fully contain the matched text so that `regex.replace_all` can apply highlight tags.

## Actual behavior

When `match_length > fragment_size / 2`, the fragment's right edge (`raw_end`) falls short of `m.end()`. The regex applied to the truncated fragment cannot find the full match, so the fragment is returned with no highlighting.

## Root cause (hypothesis)

The fragment is centered on `m.start()` with a fixed window of `fragment_size` bytes, giving the match only half the window width for forward coverage. The calculation does not account for the length of the match itself.

## Suggested fix

Ensure `raw_end` covers at least `m.end()`:

```rust
let effective_size = opts.fragment_size.max(m.end() - m.start());
let raw_start = m.start().saturating_sub(effective_size / 2);
let raw_end = usize::min(text.len(), raw_start.saturating_add(effective_size));
```

Or more conservatively, just extend the end:

```rust
let raw_end = usize::min(text.len(), raw_start.saturating_add(opts.fragment_size).max(m.end()));
```

## Environment

- Commit: fc1ff89
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes #285.

`highlight_fragments` centered its fragment on `m.start()` with a fixed width of `fragment_size`, giving the match only `fragment_size / 2` bytes of forward coverage. When a phrase match was longer than that (common for multi-term phrase queries, or phrase matches in multi-byte UTF-8), the tail of the match fell outside the fragment. The subsequent `regex.replace_all` on the truncated fragment then failed to find the full match, so the fragment was returned **without any highlight tags** — a silent correctness failure.

## Fix

Expand the window to at least the match length, and extend the end to cover `m.end()`, so `replace_all` can always apply highlight tags.

```rust
let match_len = m.end() - m.start();
let effective_size = opts.fragment_size.max(match_len);
let raw_start = m.start().saturating_sub(effective_size.saturating_sub(match_len) / 2);
let raw_end = usize::min(
  text.len(),
  raw_start.saturating_add(effective_size).max(m.end()),
);
```

The `effective_size - match_len` recentering keeps the match visually near the middle of the fragment when `fragment_size > match_len`, and degrades gracefully when the match is larger than the requested fragment (rather than pushing the fragment entirely to the left of the match).

## Test plan
- [x] Added regression test `long_phrase_match_is_fully_contained_in_fragment` reproducing the scenario from the issue (100-byte fragment, ~57-byte phrase match starting at byte ~46).
- [x] `cargo test -p searchlite-core --lib` — 373 passed.
- [x] Existing highlight tests still pass (first-term, multibyte, char-boundary snapping).

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"ff7b0b3e-bebc-4395-8d07-9f45ddff4f9d","pid":47451,"acquiredAt":1776445365687}
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"ff7b0b3e-bebc-4395-8d07-9f45ddff4f9d","pid":47451,"acquiredAt":1776445365687}
--- a/searchlite-core/src/index/highlight.rs
+++ b/searchlite-core/src/index/highlight.rs
@@ -51,7 +51,9 @@ pub fn highlight_fragments(
       // `replace_all` on the truncated fragment silently fails to highlight.
       let match_len = m.end() - m.start();
       let effective_size = opts.fragment_size.max(match_len);
-      let raw_start = m.start().saturating_sub(effective_size.saturating_sub(match_len) / 2);
+      let raw_start = m
+        .start()
+        .saturating_sub(effective_size.saturating_sub(match_len) / 2);
       let raw_end = usize::min(
         text.len(),
         raw_start.saturating_add(effective_size).max(m.end()),
@@ -152,8 +154,7 @@ mod tests {
     let text = "aaaaaaaaa bbbbbbbb cccccccc dddddddd eeeeeeee \
                 the quick brown fox jumps over the lazy dog near the river yyyy";
     let phrase: Vec<String> = [
-      "the", "quick", "brown", "fox", "jumps", "over", "the", "lazy", "dog", "near", "the",
-      "river",
+      "the", "quick", "brown", "fox", "jumps", "over", "the", "lazy", "dog", "near", "the", "river",
     ]
     .iter()
     .map(|s| s.to_string())
--- a/searchlite-core/src/index/highlight.rs
+++ b/searchlite-core/src/index/highlight.rs
@@ -47,8 +47,15 @@ pub fn highlight_fragments(
   let mut offset = 0usize;
   for _ in 0..opts.number_of_fragments {
     if let Some(m) = re.find_at(text, offset) {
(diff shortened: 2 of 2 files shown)


davidkelley:
@codex review

PR #292, opened 2026-04-17
fix(rescore): shrink sort_window by removed hits to exclude non-rescored (BUG-291)

### LINKED ISSUE #291: fix(rescore): sort_window includes non-rescored hits after min_score removal shifts window boundary
## Summary

When `rescore_hits` removes hits from within the rescore window (because `evaluate_compiled_score` returns `None` — e.g. when the rescore query uses `function_score` with `min_score`), the subsequent sort window is computed from `rescore.window_size.min(hits.len())` using the *post-removal* `hits.len()`. This pulls non-rescored hits (originally positioned outside the window) into the sort range, causing them to be re-sorted against rescored hits whose scores are on a different scale (combined BM25+rescore vs. raw BM25).

## Affected code

- `searchlite-core/src/api/reader.rs:2139` — `let sort_window = rescore.window_size.min(hits.len());`
- `searchlite-core/src/api/reader.rs:1993` — `.take(window)` limits rescoring to the original window
- `searchlite-core/src/api/reader.rs:2103-2106` — `to_remove.push(hit_idx)` when score evaluation returns `None`
- `searchlite-core/src/api/reader.rs:2132-2137` — removal shifts all later elements left

## Reproduction / evidence

**Setup:** 10 hits, `rescore.window_size = 7`.

1. `window = min(7, 10) = 7` — only `hits[0..7]` are rescored (line 1993: `.take(window)`).
2. Suppose indices 1 and 3 are removed by `to_remove` (because `evaluate_compiled_score` returns `None` — triggered by `function_score` with `min_score` in the rescore query).
3. After removal: `hits.len() = 8`. The array is now:
   - New positions 0–4: originally indices 0, 2, 4, 5, 6 — all **rescored** (combined scores)
   - New positions 5–7: originally indices 7, 8, 9 — **NOT rescored** (raw BM25 scores)
4. `sort_window = min(7, 8) = 7` — sorts `hits[0..7]`, which includes positions 5 and 6 (non-rescored).
5. Two non-rescored hits (with raw BM25 scores) are sorted against five rescored hits (with combined BM25+rescore scores). The resulting order mixes scores of different natures, producing incorrect ranking.

**Concrete example:**
- Hit A (position 0, rescored): original BM25 = 9.0, combined = 3.0 (rescore penalty)
- Hit B (position 7, not rescored): original BM25 = 5.0
- After removal shifts B into the sort window, B (5.0) sorts above A (3.0)
- But B was never rescored — its 5.0 is a raw BM25 score, not comparable to A's combined score

## Expected behavior

The sort window should only cover hits that were actually rescored. After removing `to_remove.len()` hits from the original `window`, the rescored survivors occupy positions `0..(window - to_remove.len())`. The sort should be limited to that range:

```rust
let removed = to_remove.len();
// ... removal code ...
let sort_window = (window - removed).min(hits.len());
```

This ensures non-rescored hits that shifted left due to removals are never re-sorted against rescored hits.

## Actual behavior

`sort_window` is computed from `rescore.window_size.min(hits.len())` using the post-removal length, which can exceed the number of rescored survivors and include non-rescored hits in the sort range.

## Root cause (hypothesis)

The `sort_window` calculation on line 2139 does not account for the removals that happened on lines 2132–2137. It re-derives the window from `rescore.window_size` and the new `hits.len()` instead of tracking how many rescored hits survived the removal phase.

## Suggested fix

Capture `to_remove.len()` before the removal loop drains it, and subtract from `window`:

```rust
let removed = to_remove.len();
if removed > 0 {
    to_remove.sort_unstable();
    to_remove.dedup();
    for idx in to_remove.into_iter().rev() {
        hits.remove(idx);
    }
}
let sort_window = window.saturating_sub(removed).min(hits.len());
if sort_window > 0 {
    hits[..sort_window].sort_by(|a, b| a.key.cmp(&b.key));
}
```

Note: `to_remove` is deduped before counting, so `removed` should be captured after dedup but before drain. Alternatively, compute `removed` after dedup:

```rust
to_remove.sort_unstable();
to_remove.dedup();
let removed = to_remove.len();
for idx in to_remove.into_iter().rev() {
    hits.remove(idx);
}
```

## Environment

- Commit: 681f05d
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes #291. When `rescore_hits` drops hits from inside the rescore window because `evaluate_compiled_score` returns `None` (e.g. `function_score` + `min_score` in the rescore query), the subsequent sort window was recomputed from `rescore.window_size.min(hits.len())` on the post-removal list. Hits that originally sat outside the window shifted left into the sort range and were re-sorted against rescored hits whose scores are on a different scale (combined BM25+rescore vs. raw BM25), producing incorrect ranking.

## What changed

- **`searchlite-core/src/api/reader.rs`** (`rescore_hits`): capture `to_remove.len()` after dedup, then derive `sort_window = window.saturating_sub(removed).min(hits.len())`. Only the rescored survivors (positions `0..(window - removed)`) are re-sorted; shifted non-rescored hits keep their original relative rank.

- **`searchlite-core/tests/function_score.rs`** — new regression test `rescore_sort_window_excludes_non_rescored_after_removal`. The test:
  - Scores three docs by `popularity` (doc-1=10, doc-3=5, doc-2=1).
  - Uses `window_size=2`, so doc-1 and doc-3 are rescored and doc-2 is not.
  - Rescore is a `function_score` with `field_value_factor(factor=0.001)` and `min_score=0.008`, which yields `0.01` for doc-1 (passes) and `0.005` for doc-3 (returns `None`, so doc-3 is dropped).
  - Under `RescoreMode::Multiply`, doc-1's combined score becomes `10 * 0.01 = 0.1`, while doc-2's raw score (`1.0`) is untouched.
  - Without the fix the post-removal sort re-orders doc-2 above doc-1 (`1.0 > 0.1`), even though doc-2 was never rescored. With the fix the sort range shrinks to a single element and the final order is `[doc-1, doc-2]`.

Verified the test fails on `main` and passes on this branch.

## Test plan

- [x] `cargo test -p searchlite-core --test function_score` — 20/20 pass (19 existing + 1 new regression).
- [x] `cargo test -p searchlite-core --lib` — 384/384 pass.
- [x] `cargo fmt --all -- --check` clean.

Closes #291

### DIFF
--- a/searchlite-core/src/api/reader.rs
+++ b/searchlite-core/src/api/reader.rs
@@ -2129,14 +2129,22 @@ impl IndexReader {
         }
       }
     }
-    if !to_remove.is_empty() {
+    let removed = if !to_remove.is_empty() {
       to_remove.sort_unstable();
       to_remove.dedup();
+      let n = to_remove.len();
       for idx in to_remove.into_iter().rev() {
         hits.remove(idx);
       }
-    }
-    let sort_window = rescore.window_size.min(hits.len());
+      n
+    } else {
+      0
+    };
+    // Only hits that were actually rescored (the first `window - removed`
+    // survivors) are on the combined-score scale. Any hits that shifted left
+    // due to removals still carry raw scores and must not be re-sorted against
+    // rescored survivors.
+    let sort_window = window.saturating_sub(removed).min(hits.len());
     if sort_window > 0 {
       hits[..sort_window].sort_by(|a, b| a.key.cmp(&b.key));
     }
--- a/searchlite-core/tests/function_score.rs
+++ b/searchlite-core/tests/function_score.rs
@@ -686,3 +686,73 @@ fn max_boost_caps_function_score_before_boost_mode_multiply() {
     );
   }
 }
+
+#[test]
+fn rescore_sort_window_excludes_non_rescored_after_removal() {
+  // Regression test for BUG-291: when rescore drops hits from within the
+  // window (via function_score + min_score causing evaluate_compiled_score
+  // to return None), the sort window must shrink by the number of removed
+  // hits. Otherwise non-rescored hits that shifted left into the original
+  // window bounds get re-sorted against rescored hits whose scores are on
+  // a different scale, producing incorrect ranking.
+  let reader = setup_reader();
+  // Base query: MatchAll scored by popularity (field_value_factor, Replace).
+  // Raw scores → doc-1: 10, doc-3: 5, doc-2: 1. Initial order: [doc-1, doc-3, doc-2].
+  let mut req = base_request(QueryNode::FunctionScore {
+    query: Box::new(QueryNode::MatchAll { boost: None }),
+    functions: vec![FunctionSpec::FieldValueFactor {
+      field: "popularity".into(),
+      factor: 1.0,
+      modifier: Some(FieldValueModifier::None),
+      missing: None,
+      filter: None,
+    }],
+    score_mode: Some(FunctionScoreMode::Sum),
+    boost_mode: Some(FunctionBoostMode::Replace),
+    max_boost: None,
+    min_score: None,
+    boost: None,
+  });
+  // Rescore window covers only the first two hits (doc-1 and doc-3). The
+  // rescore query yields 0.01 for doc-1 and 0.005 for doc-3; min_score=0.008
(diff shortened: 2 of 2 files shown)


davidkelley:
@codex review

---

PR #297, opened 2026-04-18
fix(aggs): accept leading sign in parse_interval_seconds (BUG-295)

### LINKED ISSUE #295: fix(aggs): parse_interval_seconds rejects negative offsets, breaking date_histogram offset: "-6h"
## Summary

`parse_interval_seconds` only accepts leading ASCII digits and `.`, so any negative duration string (e.g. `"-6h"`, `"-30m"`) returns `None`. Because the date_histogram request validator treats `None` as an invalid offset, users who specify a negative offset receive an error:

> date_histogram \`name\` offset \`-6h\` is invalid

Negative offsets are a documented Elasticsearch feature used to shift bucket boundaries backward (commonly for timezone alignment, e.g. UTC-6 via `offset: "-6h"`). Searchlite should accept them.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:4168-4192` — `parse_interval_seconds` scans the leading numeric portion with `ch.is_ascii_digit() || ch == '.'`, which does not match a leading `-`. A spec like `"-6h"` breaks out of the loop immediately, leaving `idx == 0`, and the function returns `None`.
- `searchlite-core/src/api/reader.rs:3107-3115` — The date_histogram request validator calls `parse_interval_seconds(offset)` and rejects the request when the result is `None`, surfacing the confusing "invalid" error to the caller.
- `searchlite-core/src/query/aggs/mod.rs:1438-1443` — `DateHistogramCollector::new` converts the parsed offset to `i64` millis via `(s * 1_000.0) as i64`. The downstream arithmetic (`bucket_start`, `add_calendar`) already handles negative `offset_millis` correctly (the BUG-289 fix added `checked_sub`/`checked_add` for both signs), so the only gap is in the parser.

## Reproduction / evidence

```rust
// parse_interval_seconds("-6h") traces:
//   spec = "-6h"
//   loop: '-' is not ascii_digit and not '.', break
//   idx == 0 → return None
assert_eq!(parse_interval_seconds("-6h"), None);
// Expected: Some(-21600.0)  (i.e. -6 * 3600)
```

A search request with:
```json
{
  "aggs": {
    "by_day": {
      "date_histogram": {
        "field": "timestamp",
        "fixed_interval": "1d",
        "offset": "-6h"
      }
    }
  }
}
```
returns HTTP 400 with `"date_histogram \`by_day\` offset \`-6h\` is invalid"`.

## Expected behavior

`parse_interval_seconds("-6h")` should return `Some(-21600.0)`. The date_histogram validator should accept negative offsets, and `DateHistogramCollector::new` should produce `offset_millis = -21600000`. Bucket keys should then shift backward by 6 hours, matching Elasticsearch semantics.

## Actual behavior

`parse_interval_seconds` returns `None` for any input starting with `-`. The validator rejects the request with a confusing "invalid" error. Users cannot use negative offsets at all.

## Root cause (hypothesis)

`parse_interval_seconds` was written to parse positive durations only. Its leading-character scan (`is_ascii_digit() || ch == '.'`) does not account for a leading `-` sign. The fix is to optionally consume a leading `-` before entering the digit loop, and negate the result.

## Suggested fix

In `parse_interval_seconds`, before the digit-scanning loop, check for and consume a leading `-`:

```rust
let mut chars = spec.chars().peekable();
let negative = chars.peek() == Some(&'-');
if negative { chars.next(); }
// ... scan digits from `chars` ...
let value: f64 = /* parsed */;
let value = if negative { -value } else { value };
```

## Environment

- Commit: b50f147
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes #295. `parse_interval_seconds` only scanned a leading run of `0-9` and `.`, so any spec starting with `-` exited with `idx == 0` and returned `None`. The date_histogram request validator treats a `None` offset parse as invalid, so a legitimate request like `offset: \"-6h\"` (used to shift bucket boundaries back for timezones such as UTC-6) returned HTTP 400 `offset \`-6h\` is invalid` instead of being honored.

The parser now consumes an optional leading `-` before the digit loop and negates the result. Sites that already guard against negatives stay correct:

- `api/reader.rs:3094` (`fixed_interval`) keeps its `secs > 0.0` filter, so negative fixed intervals remain rejected.
- `date_histogram_span_exceeds_cap` short-circuits when `fixed_millis <= 0`, so span checks are unaffected.
- `DateHistogramCollector::new` already stores `offset_millis` as signed `i64`, and `bucket_start` / `add_calendar` use checked arithmetic (BUG-289) that copes with either sign.

## Test plan

- [x] `cargo test -p searchlite-core --lib parse_interval_seconds` — new parser cases for negative inputs and malformed negatives.
- [x] `cargo test -p searchlite-core --test aggregation_bounds` — all 52 tests pass, including the new `fixed_interval_accepts_negative_offset` end-to-end test that verifies `offset: -6h` shifts daily boundaries back by six hours.


davidkelley:
@codex review

PR #299, opened 2026-04-18
fix(aggs): sum_bucket returns null instead of 0.0 for empty input (BUG-298)

### LINKED ISSUE #298: fix(aggs): sum_bucket returns 0.0 instead of null when no buckets contribute metric values
## Summary

`sum_bucket` unconditionally returns `Some(0.0)` (serialized as `"value": 0.0`) when no parent bucket carries a value for the referenced metric. It should return `None` (serialized with the `value` field omitted), matching both Elasticsearch semantics and the `avg_bucket` fix shipped in BUG-283 (#283).

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3217-3228` — The `SumBucket` branch initializes `sum = 0.0`, iterates over buckets, and unconditionally wraps the result in `Some(sum)`. Unlike the `AvgBucket` branch (lines 3198-3216), it does not track whether any bucket contributed a value. When no bucket has the referenced metric, the response contains `"value": 0.0` instead of omitting the field.

## Reproduction / evidence

```rust
// Setup: two buckets with no sub-aggregation matching the buckets_path.
let mut buckets = vec![
    BucketResponse { key: json!("a"), doc_count: 1, aggregations: BTreeMap::new() },
    BucketResponse { key: json!("b"), doc_count: 2, aggregations: BTreeMap::new() },
];
let mut pipeline = BTreeMap::new();
pipeline.insert(
    "total_price".to_string(),
    Aggregation::SumBucket(BucketMetricAggregation {
        buckets_path: "price_stats.sum".to_string(),
    }),
);
let responses = apply_pipeline_aggs(&pipeline, &mut buckets);

// Current: responses["total_price"] == SumBucket { value: Some(0.0) }
// Expected: responses["total_price"] == SumBucket { value: None }
```

The analogous test for `avg_bucket` (`avg_bucket_returns_none_when_all_buckets_missing_metric` at line 5822) asserts `value: None` and passes. No equivalent test exists for `sum_bucket`.

## Expected behavior

When no bucket contributes a value for the referenced metric, `sum_bucket` should return `OptionalBucketMetricResponse { value: None }`. The serialized JSON should omit the `value` field (via the existing `skip_serializing_if = "Option::is_none"` on `OptionalBucketMetricResponse`), matching Elasticsearch's `null` response for `sum_bucket` with no contributing buckets.

## Actual behavior

`sum_bucket` returns `OptionalBucketMetricResponse { value: Some(0.0) }`, serializing as `"value": 0.0`. Users cannot distinguish "no data" from "data that sums to zero."

## Root cause (hypothesis)

The `SumBucket` branch was written before `AvgBucket` received its BUG-283 fix. `AvgBucket` already had `count` tracking (for computing the average), so adding the `count == 0 → None` guard was natural. `SumBucket` never needed `count` for its computation, so the empty-input case was not addressed at the same time.

## Suggested fix

Track a boolean or counter in the `SumBucket` branch and return `None` when no values were summed:

```rust
Aggregation::SumBucket(cfg) => {
    let mut sum = 0.0_f64;
    let mut count = 0usize;
    for bucket in buckets.iter() {
        if let Some(val) = bucket_metric_value(bucket, &cfg.buckets_path) {
            sum += val;
            count += 1;
        }
    }
    let value = if count > 0 { Some(sum) } else { None };
    responses.insert(
        name.to_string(),
        AggregationResponse::SumBucket(OptionalBucketMetricResponse { value }),
    );
}
```

## Environment

- Commit: 7f90b83
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Resolves #298.

`sum_bucket` unconditionally returned `Some(0.0)` when no parent bucket carried a value for the referenced metric, serializing as `"value": 0.0`. Callers could not distinguish "no data" from "data that genuinely sums to 0". This mirrors the same class of bug that was fixed for `avg_bucket` in BUG-283 (#284).

The `SumBucket` branch in `apply_pipeline_aggs` now tracks whether any bucket contributed a value and emits `None` when none did, matching Elasticsearch semantics. `OptionalBucketMetricResponse` already marks `value` with `skip_serializing_if = "Option::is_none"`, so the empty-input case now omits the field from the JSON response.

### Changes

- `searchlite-core/src/query/aggs/mod.rs`:
  - `SumBucket` branch now tracks a `count` alongside the running sum and yields `None` when `count == 0` (mirrors the `AvgBucket` branch introduced in BUG-283).
  - Replaces `sum_bucket_returns_some_zero_when_no_buckets_contribute` (which codified the buggy behaviour) with `sum_bucket_returns_none_when_all_buckets_missing_metric`.
  - Adds `sum_bucket_returns_none_serializes_without_value_field` to lock in the JSON-omission contract.
  - Adds `sum_bucket_returns_some_sum_when_buckets_have_metric` as a positive-path regression guard.
  - Adds `sum_bucket_returns_some_zero_when_any_bucket_contributes_zero` to assert that a bucket contributing a real `0.0` still yields `Some(0.0)` (distinct from the empty-input case).

## Test plan

- [x] `cargo test -p searchlite-core --lib sum_bucket` — 4/4 pass
- [x] `cargo test -p searchlite-core --lib aggs::` — 69/69 pass, no regressions

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3216,14 +3216,17 @@ fn apply_pipeline_aggs(
       }
       Aggregation::SumBucket(cfg) => {
         let mut sum = 0.0_f64;
+        let mut count = 0usize;
         for bucket in buckets.iter() {
           if let Some(val) = bucket_metric_value(bucket, &cfg.buckets_path) {
             sum += val;
+            count += 1;
           }
         }
+        let value = if count > 0 { Some(sum) } else { None };
         responses.insert(
           name.to_string(),
-          AggregationResponse::SumBucket(OptionalBucketMetricResponse { value: Some(sum) }),
+          AggregationResponse::SumBucket(OptionalBucketMetricResponse { value }),
         );
       }
       Aggregation::Derivative(cfg) => {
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

davidkelley:
@codex review

PR #306, opened 2026-04-18
fix(aggs): accept leading `+` sign in parse_interval_seconds (BUG-305)

### LINKED ISSUE #305: fix(aggs): parse_interval_seconds rejects leading "+" sign, breaking date_histogram offset: "+6h"
## Summary

`parse_interval_seconds` accepts a leading `-` sign (added in the BUG-295 fix) but does not accept a leading `+` sign. Any duration string with an explicit positive prefix (e.g. `"+6h"`, `"+30m"`) returns `None`, causing the date_histogram request validator to reject the offset with a confusing error. Elasticsearch time-unit parsing accepts both `+` and `-` prefixes, so users migrating from Elasticsearch or following its documentation will hit this.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:4228-4231` — `parse_interval_seconds` strips a leading `-` via `strip_prefix('-')` but has no equivalent arm for `+`. A spec like `"+6h"` falls through to the default `(false, spec)` branch, where the loop at line 4233 immediately breaks on `'+'` (not a digit or `'.'`), leaving `idx == 0`, and the function returns `None`.
- `searchlite-core/src/api/reader.rs:3108` — The date_histogram request validator calls `parse_interval_seconds(offset)` and rejects the request when the result is `None`.
- The same parser is used for `fixed_interval` (reader.rs:3093) and histogram `interval` (reader.rs:3189, 3202), so `"+1h"` would also be rejected there.

## Reproduction / evidence

```rust
// parse_interval_seconds("+6h") traces:
//   spec = "+6h"
//   strip_prefix('-') → None → (negative=false, rest="+6h")
//   loop: '+' is not ascii_digit and not '.', break immediately
//   idx == 0 → return None
assert_eq!(parse_interval_seconds("+6h"), None);

// Expected:
assert_eq!(parse_interval_seconds("+6h"), Some(21_600.0));
```

A search request with:
```json
{
  "aggs": {
    "by_day": {
      "date_histogram": {
        "field": "timestamp",
        "calendar_interval": "day",
        "offset": "+6h"
      }
    }
  }
}
```
returns error: `date_histogram 'by_day' offset '+6h' is invalid`.

## Expected behavior

`parse_interval_seconds("+6h")` should return `Some(21_600.0)`, matching the behavior of `parse_interval_seconds("6h")`. The explicit `+` prefix is semantically identical to no prefix (positive duration) and should be stripped before numeric parsing, mirroring the existing `-` handling.

## Actual behavior

`parse_interval_seconds("+6h")` returns `None`. The date_histogram validator surfaces this as an opaque "invalid" error to the caller.

## Root cause (hypothesis)

The BUG-295 fix (commit `0eca900`) added `-` sign handling to `parse_interval_seconds` but did not add a parallel `strip_prefix('+')` arm. The commit title ("accept leading sign") suggests both signs were intended, but only the negative case was implemented and tested.

## Suggested fix

Add a `strip_prefix('+')` branch alongside the existing `-` branch:

```rust
let (negative, rest) = if let Some(tail) = spec.strip_prefix('-') {
    (true, tail)
} else if let Some(tail) = spec.strip_prefix('+') {
    (false, tail)
} else {
    (false, spec)
};
```

## Environment

- Commit: 87f5103
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes #305. `parse_interval_seconds` accepted a leading `-` (added in BUG-295) but had no parallel `+` arm, so explicit-positive duration specs such as `"+6h"` or `"+30m"` fell through to the default branch and returned `None`. That surfaced as `date_histogram 'by_day' offset '+6h' is invalid` for requests that mirror Elasticsearch's accepted syntax (ES accepts both `+` and `-` prefixes on time-unit specs).

The fix adds a second `strip_prefix('+')` branch alongside the existing `-` arm. `+` is treated as a no-op sign — numeric scan and unit lookup are unchanged, so malformed specs (`"+"`, `"++5h"`, `"+-5h"`, `"-+5h"`, `"+5x"`) continue to return `None`.

## Changes

- `searchlite-core/src/query/aggs/mod.rs` — extend the leading-sign match with an explicit `+` arm; updated the adjacent doc comment to reflect that both signs are now accepted.
- `searchlite-core/src/query/aggs/mod.rs` (tests) — add `parse_interval_seconds_accepts_explicit_positive_sign` and `parse_interval_seconds_rejects_malformed_positives`, mirroring the BUG-295 coverage.
- `searchlite-core/tests/aggregation_bounds.rs` — add `fixed_interval_accepts_explicit_positive_offset` regression in `bug_030`, parallel to the existing `fixed_interval_accepts_negative_offset`, exercising the full request path (`offset: "+6h"` with `fixed_interval: 1d`).

## Test plan

- [x] `cargo test -p searchlite-core --lib parse_interval_seconds` — all six unit tests pass (including existing negative/malformed coverage).
- [x] `cargo test -p searchlite-core --test aggregation_bounds fixed_interval_accepts` — both negative (BUG-295) and positive (BUG-305) integration regressions pass.
- [x] `cargo fmt --all --check` — clean.

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -4220,14 +4220,20 @@ pub(crate) fn parse_date(value: &str) -> Option<f64> {
 }
 
 pub(crate) fn parse_interval_seconds(spec: &str) -> Option<f64> {
-  // Accept an optional leading `-` so negative durations such as "-6h"
-  // parse correctly. Callers that only permit positive durations
+  // Accept an optional leading sign so negative durations such as "-6h"
+  // and explicitly-positive durations such as "+6h" parse correctly.
+  // Elasticsearch time-unit parsing accepts both `+` and `-` prefixes,
+  // so users migrating from Elasticsearch or following its documentation
+  // expect both. Callers that only permit positive durations
   // (e.g. `fixed_interval`) are responsible for rejecting negatives
   // via their own `secs > 0.0` guard; callers that accept negatives
   // (date_histogram `offset`) rely on this path.
-  let (negative, rest) = match spec.strip_prefix('-') {
-    Some(tail) => (true, tail),
-    None => (false, spec),
+  let (negative, rest) = if let Some(tail) = spec.strip_prefix('-') {
+    (true, tail)
+  } else if let Some(tail) = spec.strip_prefix('+') {
+    (false, tail)
+  } else {
+    (false, spec)
   };
   let mut idx = 0usize;
   for ch in rest.chars() {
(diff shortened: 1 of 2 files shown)


davidkelley:
@codex review

davidkelley:
@codex review

PR #314, opened 2026-04-18
fix(aggs): handle unary `+` in bucket_script tokenizer (BUG-313)

### LINKED ISSUE #313: fix(aggs): bucket_script tokenizer treats unary `+` as binary Add, silently producing None for valid expressions
## Summary

The `bucket_script` tokenizer in `aggs/mod.rs` treats a unary `+` as a binary `Add` operator, causing `eval_rpn` to underflow the stack and return `None` for any expression that uses a leading or inline `+` sign. The BUG-309 fix (commit 463f705) corrected this exact issue in the `script_score` tokenizer (`script.rs`), but the separate `tokenize_script` function used by `bucket_script` was not updated.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3500-3509` — the `'+' | '-' | '*' | '/'` match arm only checks for unary `-` (`ch == '-' && expect_unary`), not unary `+`. When `expect_unary = true` and `ch = '+'`, the code falls through to push a binary `ScriptToken::Op('+')`.
- `searchlite-core/src/query/script.rs:278-298` — the analogous `script_score` tokenizer, which now correctly handles unary `+` (fixed by BUG-309).

## Reproduction / evidence

Given a `bucket_script` with the expression `+5 * 2` (or `+a`, or `a * +b`):

**Tokenization trace** for `+5 * 2`:
1. `ch = '+'`, `expect_unary = true`. The unary guard (`ch == '-' && expect_unary`) is false because `ch` is `+`, not `-`. Falls through to push `ScriptToken::Op('+')` as a binary Add.
2. `ch = '5'` → `Number(5.0)`
3. `ch = '*'` → `Op('*')`
4. `ch = '2'` → `Number(2.0)`

Tokens: `[Op('+'), Number(5.0), Op('*'), Number(2.0)]`

**RPN conversion** via `to_rpn`: `[Number(5.0), Number(2.0), Op('*'), Op('+')]`

**Evaluation** via `eval_rpn`:
1. Push 5.0, push 2.0
2. `Op('*')`: pop 2.0, pop 5.0, push 10.0
3. `Op('+')`: pop 10.0, pop → **stack empty** → returns `None`

Similarly:
- `eval_bucket_script("+a", &vars)` → `None` (expected: value of `a`)
- `eval_bucket_script("a + +b", &vars)` → `None` (expected: `a + b`)
- `eval_bucket_script("(+3) * 2", &vars)` → `None` (expected: `6.0`)

No existing test covers unary `+` in `bucket_script` — tests exist only for unary negation.

## Expected behavior

Unary `+` should be treated as a no-op (identity operator), consistent with the `script_score` tokenizer after BUG-309. Expressions like `+5 * 2` should evaluate to `10.0`, `+a` should return the value of `a`, and `a * +b` should return `a * b`.

## Actual behavior

Any `bucket_script` expression containing a unary `+` silently returns `None`, producing no value for that bucket. This means the `bucket_script` aggregation result is missing for affected buckets, and any downstream pipeline aggregation referencing it will also produce no value.

## Root cause (hypothesis)

The BUG-309 fix targeted `tokenize` in `script.rs` (the `script_score` path) but did not update `tokenize_script` in `aggs/mod.rs` (the `bucket_script` path). The two tokenizers have nearly identical structure and share the same design — the `bucket_script` tokenizer was likely written before BUG-309 and inherited the same unary-plus oversight, but was not included in the BUG-309 fix scope.

## Suggested fix

In `searchlite-core/src/query/aggs/mod.rs`, update the operator match arm in `tokenize_script` to handle unary `+` the same way the `script_score` tokenizer does:

When `expect_unary = true` and `ch = '+'`:
- If the next character is a digit or `.`, consume it and parse a positive number literal
- Otherwise, skip the `+` entirely (unary plus is a no-op) and leave `expect_unary` true

## Environment

- Commit: 463f705
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes #313.

The `tokenize_script` function in `searchlite-core/src/query/aggs/mod.rs` only treated unary `-` as a sign. A leading `+` fell through to the binary-operator arm and emitted `ScriptToken::Op('+')`. During RPN evaluation this popped an operand that was never pushed, so `eval_bucket_script` silently returned `None` for any expression that used a unary `+` (for example `+5 * 2`, `+a`, `a * +b`, or `(+3) * 2`).

This mirrors the BUG-309 fix applied to the `script_score` tokenizer in `script.rs`:

- The numeric-literal branch now also consumes a leading unary `+` (the `+` is dropped so the parser still sees a plain positive literal).
- A bare unary `+` before a variable or `(` in the operator match arm is now treated as a no-op — it is consumed and `expect_unary` stays true so the next token is parsed as an operand.

Unary `+` now behaves as an identity operator, consistent with the `script_score` path.

## Test plan

- [x] `cargo test --package searchlite-core --lib bucket_script` — all 24 tests pass, including six new regression tests:
  - `bucket_script_unary_plus_on_number_literal` (`+5 * 2` → `10.0`)
  - `bucket_script_unary_plus_on_variable` (`+a` → `a`)
  - `bucket_script_binary_plus_followed_by_unary_plus` (`a + +b` → `a + b`)
  - `bucket_script_unary_plus_multiplication_with_variable` (`a * +b` → `a * b`)
  - `bucket_script_unary_plus_in_parens` (`(+3) * 2` → `6.0`)
  - `bucket_script_unary_plus_before_parenthesized_expression` (`+(a + b)` → `a + b`)
- [x] `cargo test --package searchlite-core --lib` — all 418 tests pass (no regression).
- [x] `cargo clippy --package searchlite-core --lib --tests -- -D warnings` — clean.

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3455,7 +3455,7 @@ fn tokenize_script(script: &str) -> Option<Vec<ScriptToken>> {
     let looks_numeric = ch.is_ascii_digit()
       || ch == '.'
       || (expect_unary
-        && ch == '-'
+        && (ch == '-' || ch == '+')
         && chars
           .clone()
           .nth(1)
@@ -3465,6 +3465,10 @@ fn tokenize_script(script: &str) -> Option<Vec<ScriptToken>> {
       let mut num = String::new();
       if ch == '-' {
         num.push('-');
+        chars.next();
+      } else if ch == '+' {
+        // Unary plus is a no-op; consume the `+` without adding it to the
+        // literal so the parser sees a plain positive number.
         chars.next();
       }
       while let Some(next) = chars.peek() {
@@ -3501,6 +3505,13 @@ fn tokenize_script(script: &str) -> Option<Vec<ScriptToken>> {
       '+' | '-' | '*' | '/' => {
         if ch == '-' && expect_unary {
           tokens.push(ScriptToken::Neg);
+          chars.next();
+          continue;
+        }
+        if ch == '+' && expect_unary {
+          // Unary plus before a non-numeric operand (variable or `(`) is a
+          // no-op. Consume the `+` and leave `expect_unary` true so the next
+          // token is still parsed as an operand.
           chars.next();
           continue;
         }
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

davidkelley:
Rebased onto latest main.

@codex review

PR #318, opened 2026-04-18
fix(query): apply max_expansions globally across segments (BUG-316)

### LINKED ISSUE #316: fix(query): expand_prefix/wildcard/regex apply max_expansions per-segment instead of globally, exceeding configured limit on multi-segment indexes
## Summary

`expand_prefix`, `expand_wildcard`, and `expand_regex` in `term_expansion.rs` reset their `expanded` counter for each segment, applying `max_expansions` per-segment rather than as a global limit across all segments. On a multi-segment index this allows the total number of expanded terms to exceed the configured `max_expansions` by up to `(num_segments - 1) * max_expansions`, producing more term matches than the user specified and causing slower-than-expected queries.

## Affected code

- `searchlite-core/src/api/term_expansion.rs:372-399` — `expand_prefix`: `let mut expanded = 0usize` is inside the `for seg in segments.iter()` loop (line 373), so it resets to 0 for each segment.
- `searchlite-core/src/api/term_expansion.rs:445-476` — `expand_wildcard`: same per-segment counter pattern (line 446).
- `searchlite-core/src/api/term_expansion.rs:799-831` — `expand_regex`: same per-segment counter pattern (line 800).
- `searchlite-core/src/api/term_expansion.rs:888-927` — `expand_term_fuzzy`: correctly uses a **global** counter (`let mut expansions = 0usize` at line 888, outside the segment loop) with a labeled `break 'segments` to enforce the limit across all segments.

## Reproduction / evidence

Consider an index with 3 segments, each containing unique terms starting with "ap":
- Segment 0: ["app", "apple"]
- Segment 1: ["apply", "apt"]
- Segment 2: ["apricot"]

A prefix query for `"ap"` with `max_expansions = 2`:

**Expected (global limit):** 2 unique terms total (e.g. "app", "apple"), matching fuzzy expansion behavior.

**Actual (per-segment limit):** Up to 5 unique terms — each segment contributes up to 2 unique terms independently because `expanded` resets to 0 at the start of each segment loop iteration.

Trace through `expand_prefix`:
1. Segment 0: `expanded=0` → "app" (insert, expanded=1) → "apple" (insert, expanded=2) → **break** (inner loop only)
2. Segment 1: `expanded=0` (reset!) → "apply" (insert, expanded=1) → "apt" (insert, expanded=2) → **break**
3. Segment 2: `expanded=0` (reset!) → "apricot" (insert, expanded=1)

Result: 5 expanded terms, not 2.

The `seen` HashSet deduplicates terms across segments but does not prevent the counter from re-accumulating unique terms. In the worst case (all segments have disjoint terms), the total is `num_segments * max_expansions`.

## Expected behavior

`max_expansions` should limit the total number of unique expanded terms across all segments, consistent with:
1. The fuzzy expansion path in the same file (`expand_term_fuzzy`), which correctly uses a global counter.
2. Elasticsearch semantics, where `max_expansions` is a global cap on rewritten terms.

## Actual behavior

The per-segment counter allows the total to exceed `max_expansions` on multi-segment indexes. Single-segment indexes are unaffected.

## Root cause (hypothesis)

The three expansion functions (`expand_prefix`, `expand_wildcard`, `expand_regex`) were written with the counter inside the segment loop, while `expand_term_fuzzy` was written (or later corrected) with the counter outside the loop. The inconsistency was never caught because single-segment test indexes do not expose the difference, and the `seen` HashSet masking duplicates makes the over-expansion less visible.

## Suggested fix

Move the `expanded` counter outside the segment loop and use a labeled break, matching the pattern in `expand_term_fuzzy`:

```rust
let mut expanded = 0usize;
'segments: for seg in segments.iter() {
    for key in seg.terms_with_prefix(&prefix_key) {
        if expanded >= max_expansions {
            break 'segments;
        }
        // ...
        if !seen.insert(key.to_owned()) {
            continue;
        }
        // ...
        expanded += 1;
    }
}
```

Apply the same change to all three functions (`expand_prefix`, `expand_wildcard`, `expand_regex`).

## Environment

- Commit: b03dcbc
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- `expand_prefix`, `expand_wildcard`, and `expand_regex` were resetting their `expanded` counter inside the segment loop, so `max_expansions` was enforced per-segment instead of globally. On a multi-segment index the total number of expanded terms could reach `num_segments * max_expansions`, exceeding the configured cap.
- Hoist the counter outside the segment loop and use a labeled `break 'segments` to enforce the limit globally, matching the pattern already used by `expand_term_fuzzy`.
- Adds multi-segment regression tests for all three expansion paths.

Fixes #316.

## Test plan

- [x] `cargo test -p searchlite-core --lib cap_is_global_across_segments` — new tests fail without the fix and pass with it.
- [x] `cargo test -p searchlite-core --lib expansion` — all existing prefix/wildcard/regex/fuzzy expansion tests still pass.
- [x] `cargo clippy -p searchlite-core --lib --tests` — clean.
- [x] `cargo fmt -p searchlite-core --check` — clean.

### DIFF
--- a/searchlite-core/src/api/reader.rs
+++ b/searchlite-core/src/api/reader.rs
@@ -4623,6 +4623,240 @@ mod tests {
     );
   }
 
+  /// BUG-316 regression: `max_expansions` must bound the total number of
+  /// expanded terms across *all* segments, not per-segment. Before the fix
+  /// the `expanded` counter in `expand_prefix`/`expand_wildcard`/`expand_regex`
+  /// lived inside the segment loop and reset to 0 for each segment, letting
+  /// a 3-segment index return up to `3 * max_expansions` unique terms.
+  #[test]
+  fn prefix_expansion_cap_is_global_across_segments() {
+    let dir = tempfile::tempdir().unwrap();
+    let path = dir.path().join("idx-prefix-multi-seg");
+    let idx = Index::create(
+      &path,
+      Schema::default_text_body(),
+      IndexOptions {
+        path: path.clone(),
+        create_if_missing: true,
+        enable_positions: true,
+        bm25_k1: 0.9,
+        bm25_b: 0.4,
+        storage: StorageType::Filesystem,
+        #[cfg(feature = "vectors")]
+        vector_defaults: None,
+      },
+    )
+    .unwrap();
+    // Three segments, each containing disjoint unique terms starting with "ap".
+    for batch in [
+      vec![("1", "app"), ("2", "apple")],
+      vec![("3", "apply"), ("4", "apt")],
+      vec![("5", "apricot")],
+    ] {
+      let mut writer = idx.writer().unwrap();
+      for (id, body) in batch {
+        writer
+          .add_document(&Document {
+            fields: [
+              ("_id".into(), serde_json::json!(id)),
+              ("body".into(), serde_json::json!(body)),
+            ]
+            .into_iter()
+            .collect(),
+          })
+          .unwrap();
+      }
+      writer.commit().unwrap();
+    }
+    let reader = idx.reader().unwrap();
+    assert!(
+      reader.segments.len() >= 2,
+      "expected multiple segments, got {}",
+      reader.segments.len()
+    );
+    let default_fields: Vec<String> = reader
+      .manifest
+      .schema
+      .text_fields
+      .iter()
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

PR #321, opened 2026-04-18
fix(aggs): numeric tiebreaker in bucket_sort_cmp for numeric bucket keys (BUG-320)

### LINKED ISSUE #320: fix(aggs): bucket_sort_cmp tiebreaker uses lexicographic string comparison on numeric bucket keys, producing wrong order when explicit sort specs tie
## Summary

`bucket_sort_cmp` falls back to a lexicographic string comparison of bucket keys when all explicit sort specs compare equal. For `histogram` and `date_histogram` parent aggregations whose bucket keys are numeric (JSON numbers), this tiebreaker produces an incorrect ordering — the same class of bug that BUG-296 fixed for the `_key` sort spec, but in the residual tiebreaker that BUG-296 did not address.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3685` — the fallback `bucket_key_string(&a.key).cmp(&bucket_key_string(&b.key))` at the end of `bucket_sort_cmp`. This path is reached when every `BucketSortSpec` in `specs` compares equal (or when `specs` is empty).
- `searchlite-core/src/query/aggs/mod.rs:2785-2791` — `bucket_key_string` converts any JSON value to a `String` via `key.to_string()`, losing numeric ordering semantics.

## Reproduction / evidence

Index documents with a numeric field (e.g. `temperature`) spanning negative values, and run a `histogram` aggregation with a `bucket_sort` sub-aggregation sorted by `_count`:

```json
{
  "aggs": {
    "by_temp": {
      "histogram": { "field": "temperature", "interval": 10 },
      "aggs": {
        "sorted": {
          "bucket_sort": {
            "sort": [{ "_count": "asc" }],
            "size": 5
          }
        }
      }
    }
  }
}
```

Suppose the histogram produces buckets with keys `-30`, `-20`, `-10`, `0`, `10`, `20`, `30`, all having the same `doc_count` (e.g. 5). The `_count` comparison is `Equal` for every pair, so `bucket_sort_cmp` falls through to line 3685.

`bucket_key_string` converts JSON number `-10` to `"-10"`, `-20` to `"-20"`, `-30` to `"-30"`, etc. Lexicographic comparison yields:

```
"-10" < "-20" < "-30" < "0" < "10" < "20" < "30"
```

The correct numeric tiebreaker order is:

```
-30 < -20 < -10 < 0 < 10 < 20 < 30
```

With `size: 5`, the user receives buckets `{-10, -20, -30, 0, 10}` instead of the expected `{-30, -20, -10, 0, 10}`.

The same issue arises with an empty sort list (`"sort": []`), where the tiebreaker is the *only* comparison — all buckets are reordered lexicographically, breaking the parent aggregation's natural key order.

## Expected behavior

The tiebreaker should use the same numeric-aware comparison introduced by BUG-296 for the `_key` sort spec — i.e., compare integer keys as `i128`, float keys via `total_cmp`, and only fall back to string comparison for string-typed keys.

## Actual behavior

The tiebreaker converts all keys (including numeric ones) to strings via `bucket_key_string` and compares lexicographically. This misordering is observable whenever:
1. All explicit `bucket_sort` specs tie (e.g. equal `_count`), or
2. `sort` is an empty list (pagination-only use case).

## Root cause (hypothesis)

BUG-296 (PR #300, commit `9e2b51f`) introduced `BucketSortComparable` with `I128`/`F64`/`Str` variants and wired them into `bucket_sort_value` for the `"key" | "_key"` sort spec at lines 3691-3703. However, the residual tiebreaker on line 3685 — which resolves ties when all *explicit* specs compare equal — was left unchanged and still uses the pre-BUG-296 string path. The tiebreaker is effectively a "hidden `_key` sort" that bypasses the numeric-aware comparator.

## Suggested fix

Replace the string tiebreaker on line 3685 with a numeric-aware comparison that reuses the `BucketSortComparable` machinery. For example:

```rust
fn bucket_key_comparable(key: &serde_json::Value) -> BucketSortComparable {
    if let Some(n) = key.as_i64() {
        BucketSortComparable::I128(n as i128)
    } else if let Some(n) = key.as_u64() {
        BucketSortComparable::I128(n as i128)
    } else if let Some(n) = key.as_f64() {
        BucketSortComparable::F64(n)
    } else {
        BucketSortComparable::Str(bucket_key_string(key))
    }
}

// In bucket_sort_cmp, replace line 3685:
compare_sort_values(
    &bucket_key_comparable(&a.key),
    &bucket_key_comparable(&b.key),
    SortOrder::Asc,
)
```

## Environment

- Commit: 071c648
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes [#320](https://github.com/davidkelley/searchlite/issues/320).

The residual tiebreaker at the end of `bucket_sort_cmp` — reached when every explicit `BucketSortSpec` compares equal, or when `specs` is empty (pagination-only use case) — fell back to `bucket_key_string` and compared keys lexicographically. For `histogram` / `date_histogram` parents whose bucket keys are JSON numbers, this produced an incorrect ordering (`"-10" < "-20" < "-30" < "0"` lexicographically instead of `-30 < -20 < -10 < 0` numerically).

BUG-296 (PR #300) fixed this same class of bug for the explicit `_key` sort spec by introducing `BucketSortComparable::{I128, F64, Str}`, but the hidden tiebreaker on line 3685 was left on the pre-BUG-296 string path.

## Fix

- Extract the key→`BucketSortComparable` conversion from the `_key` branch into a shared `bucket_key_comparable` helper.
- Call that helper from the residual tiebreaker (via `compare_sort_values` with `SortOrder::Asc`) so the tiebreaker uses the same numeric-aware comparator as an explicit `_key` sort.
- String keys (terms aggregations) still fall back to lexicographic order.

## Test plan

Added four regression tests in `searchlite-core/src/query/aggs/mod.rs`:

- `bucket_sort_cmp_tiebreaker_orders_negative_numeric_keys_numerically` — keys `-30..30` step 10 with equal `_count`; lexicographic would interleave negatives.
- `bucket_sort_cmp_empty_specs_orders_numeric_keys_numerically` — pagination-only case (empty `sort` list) with `f64` keys.
- `bucket_sort_cmp_tiebreaker_preserves_i64_precision_above_2_pow_53` — nanosecond-timestamp keys that would collapse under `f64`.
- `bucket_sort_cmp_tiebreaker_preserves_string_key_ordering` — ensures terms aggregation string keys still sort lexicographically.

- [x] `cargo test -p searchlite-core --lib bucket_sort` — 8/8 pass
- [x] `cargo test -p searchlite-core --lib query::aggs` — 98/98 pass
- [x] `cargo fmt --check -p searchlite-core`

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3682,25 +3682,38 @@ fn bucket_sort_cmp(a: &BucketResponse, b: &BucketResponse, specs: &[BucketSortSp
       return ord;
     }
   }
-  bucket_key_string(&a.key).cmp(&bucket_key_string(&b.key))
+  // Residual tiebreaker — a hidden `_key` sort applied when every explicit
+  // spec ties (or when `specs` is empty for pagination-only use cases). Reuse
+  // the same numeric-aware comparator as the `_key` spec so numeric bucket
+  // keys (histogram/date_histogram) are not collapsed to lexicographic order
+  // (BUG-320).
+  compare_sort_values(
+    &bucket_key_comparable(&a.key),
+    &bucket_key_comparable(&b.key),
+    SortOrder::Asc,
+  )
+}
+
+// Shared by the `_key` sort spec and the residual tiebreaker in
+// `bucket_sort_cmp`. Preserve exact integer precision before falling back to
+// f64, so i64/u64 keys above 2^53 (e.g. nanosecond timestamps) compare
+// correctly.
+fn bucket_key_comparable(key: &serde_json::Value) -> BucketSortComparable {
+  if let Some(n) = key.as_i64() {
+    BucketSortComparable::I128(n as i128)
+  } else if let Some(n) = key.as_u64() {
+    BucketSortComparable::I128(n as i128)
+  } else if let Some(n) = key.as_f64() {
+    BucketSortComparable::F64(n)
+  } else {
+    BucketSortComparable::Str(bucket_key_string(key))
+  }
 }
 
 fn bucket_sort_value(bucket: &BucketResponse, spec: &BucketSortSpec) -> BucketSortComparable {
   match spec.field.as_str() {
     "_count" => BucketSortComparable::F64(bucket.doc_count as f64),
-    "key" | "_key" => {
-      // Preserve exact integer precision before falling back to f64, so
-      // i64/u64 keys above 2^53 (e.g. nanosecond timestamps) compare correctly.
-      if let Some(n) = bucket.key.as_i64() {
-        BucketSortComparable::I128(n as i128)
-      } else if let Some(n) = bucket.key.as_u64() {
-        BucketSortComparable::I128(n as i128)
-      } else if let Some(n) = bucket.key.as_f64() {
-        BucketSortComparable::F64(n)
-      } else {
-        BucketSortComparable::Str(bucket_key_string(&bucket.key))
-      }
-    }
+    "key" | "_key" => bucket_key_comparable(&bucket.key),
     path => bucket_metric_value(bucket, path)
       .map(BucketSortComparable::F64)
       .unwrap_or(BucketSortComparable::Missing),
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

---

PR #341, opened 2026-04-19
fix(vectors): reject non-finite components in query vector (BUG-340)

### LINKED ISSUE #340: fix(vectors): query vector components are not validated for finitude before normalization and similarity
## Summary

`IndexReader::plan_vectors` in `searchlite-core/src/api/reader.rs` accepts the caller-provided `VectorQuery::vector` (`Vec<f32>`) verbatim and passes it through `normalize_in_place` (for the `Cosine` metric) and then into HNSW `search`/`metric_similarity` without validating that each component is finite. `serde_json` deserializes an over-range JSON number such as `1e40` to a finite `f64` and narrows it to `f32` via the default `Deserialize for f32` impl (`visit_f64(v) -> Ok(v as f32)`), which saturates to `±f32::INFINITY`. A JSON token of `NaN`/`Infinity` is not valid JSON, but a non-finite `f32` is also trivially reachable from any caller constructing the `VectorQuery` struct directly (FFI, WASM, in-process Rust).

The non-finite component then flows through:

- `normalize_in_place` (`vectors/mod.rs:74-81`): for a single `±INF` component, `norm = (Σ vᵢ²).sqrt() = INF`, then `v /= INF` yields `0.0` for every finite component and `NaN` for the offending one — producing an all-zero / one-`NaN` normalized query.
- `metric_similarity` for `Cosine` (`vectors/mod.rs:107-120`): the dot product over a `NaN` component is `NaN`; the existing `if dot.is_nan() { 0.0 }` guard masks that specific shape, but a `+INF` query component (for Cosine, only reachable if `normalize_in_place` never executes because the metric is not Cosine — see L2 below) yields a `NaN` once multiplied by a doc component of the opposite sign.
- `metric_similarity` for `L2` (`vectors/mod.rs:118`): no normalization is performed, so the raw `±INF` component persists; `l2_distance` then computes `(INF - doc_val)² = INF`, `sum = INF`, `sqrt(INF) = INF`, and `-l2_distance` returns `-INF` with no guard.
- `HnswGraph::search` (`vectors/hnsw.rs`): neighbor selection uses `similarity` / `similarity_between` against the query vector; a non-finite distance corrupts the priority-queue ordering for the duration of the search and breaks k-NN recall even for other queries that share the same reader instance (the queue state is transient, but the per-search output order is wrong).
- `hit.vector_score` in the `SearchHit` payload (`reader.rs`): the non-finite similarity leaks into the response — `serde_json` cannot emit `NaN` / `Infinity` as JSON numbers, so the entire response serialization fails.
- Hybrid blending (`blend_scores` in `vectors/mod.rs:122-129`): `alpha * bm25 + (1.0 - alpha) * vec_component` where `vec_component = -INF` yields `-INF`. The BUG-328 guard in `compute_hybrid_score` catches this downstream, but only on the hybrid path — not on a pure vector query.

## Affected code

- `searchlite-core/src/api/reader.rs:490` — `let mut query_vec = vector_query.vector.clone();` clones the caller's vector with no finitude check on components.
- `searchlite-core/src/api/reader.rs:495` — `normalize_in_place(&mut query_vec);` runs on the unvalidated vector when the metric is `Cosine`, producing `NaN` components on out-of-range inputs.
- `searchlite-core/src/api/reader.rs:534` — `vector: query_vec,` stores the unvalidated vector on `VectorClausePlan` for later use by HNSW and metric_similarity.

The adjacent finitude guards in this function already validate the scalar knobs:

- `alpha` is checked: `if !(0.0..=1.0).contains(&alpha) || !alpha.is_finite() { bail!("vector alpha must be a finite value between 0 and 1 inclusive"); }` (line 498).
- `boost` is checked: `if boost < 0.0 || !boost.is_finite() { bail!("vector boost must be finite and non-negative"); }` (line 527).

Only the vector components themselves slip through.

## Reproduction / evidence

**L2 metric, pure vector query:**

1. Schema with a single vector field `embedding`, `metric: "l2"`, `dim: 4`.
2. Index a normal doc: `{"embedding": [0.1, 0.2, 0.3, 0.4]}`.
3. Search with `vector_query: {"field": "embedding", "vector": [1e40, 0.0, 0.0, 0.0]}` — `1e40` is valid JSON and deserializes to `f32::INFINITY`.
4. `metric_similarity(L2, [INF, 0, 0, 0], [0.1, 0.2, 0.3, 0.4]) = -INF`.
5. `hit.vector_score = -INF`; serializing the response with `serde_json::to_string` errors (`NaN`/`Infinity` are not valid JSON numbers).

**Cosine metric, pure vector query:**

1. Same schema with `metric: "cosine"`.
2. Search with `vector: [1e40, 0.0, 0.0, 0.0]`.
3. `normalize_in_place` computes `norm = INF.sqrt() = INF`, then `v /= INF` yields `[NaN, 0, 0, 0]` (`INF / INF = NaN`).
4. `metric_similarity(Cosine, [NaN, 0, 0, 0], doc)`'s `dot.is_nan()` guard returns `0.0`, so the response at least serializes, but every doc has score `0.0` — the guard hides the symptom rather than surfacing the bad input.

**NaN component:**

1. Any caller that constructs `VectorQuery { vector: vec![f32::NAN, ...], .. }` (FFI, WASM, in-process) passes `NaN` through unchecked.
2. For L2: `(NaN - doc_val)² = NaN`, `sum = NaN`, `sqrt(NaN) = NaN`, `-NaN = NaN` — not caught by any guard, leaks into `hit.vector_score` and breaks JSON serialization.
3. For Cosine: normalization is `v /= NaN = NaN`; the dot-product guard masks the symptom.

Compare with the guarded sibling paths in this codebase:

- `collect_vector_value` (index-time, BUG-330 / PR #331): rejects non-finite components at insert time with `bail!("vector field {field} contains non-finite component")`.
- `eval_rpn` (bucket_script, BUG-287): rejects non-finite results.
- `combine_function_scores` (function_score, BUG-315): rejects non-finite combined scores.
- `compute_hybrid_score` (BUG-328): drops candidate on non-finite blend.

The write-side is covered by BUG-330; the query/read-side equivalent lacks the same guard.

## Expected behavior

A non-finite component in `VectorQuery::vector` should cause `search` to return a clear error at plan-time, matching the surrounding input-validation `bail!` pattern for `alpha` and `boost`. This keeps the contract symmetric with the write side (BUG-330) — invalid inputs are rejected at the boundary rather than silently corrupting downstream similarity computations, HNSW traversal, or response serialization.

## Actual behavior

`plan_vectors` accepts the unvalidated vector, normalizes it (for Cosine, producing `NaN` components), stores it on `VectorClausePlan`, and feeds it into HNSW search and `metric_similarity`. For L2 the raw `±INF` persists into `hit.vector_score` and breaks JSON serialization of the response; for Cosine the downstream `dot.is_nan()` guard silently coerces the result to `0.0`, hiding the malformed input.

## Root cause (hypothesis)

When the vector-query validation was added in `plan_vectors`, `alpha`, `boost`, `k`, `candidate_size`, and dimension were each given their own guard, but `vector_query.vector` itself was assumed to contain caller-validated components. The assumption breaks whenever the caller is an external process deserializing JSON (which can legally carry out-of-range finite `f64` literals that saturate to `±INF` on the default `Deserialize for f32`) or a direct Rust/FFI/WASM caller constructing the struct. BUG-330 / PR #331 later added the same-class guard on the write side (`collect_vector_value`), but the read side was not updated alongside it.

## Suggested fix

Validate each component immediately after the clone, before `normalize_in_place` / storage on `VectorClausePlan`, matching the surrounding `bail!` pattern and the write-side policy in `collect_vector_value`:

```rust
let mut query_vec = vector_query.vector.clone();
for (idx, v) in query_vec.iter().enumerate() {
  if !v.is_finite() {
    bail!(
      "vector query for field `{}` contains non-finite component at index {}",
      vector_query.field,
      idx
    );
  }
}
if matches!(schema_field.metric, crate::index::manifest::VectorMetric::Cosine) {
  normalize_in_place(&mut query_vec);
}
```

This rejects all three classes:

- Finite `f64` magnitude > `f32::MAX` saturating to `±f32::INFINITY` on JSON→`f32` narrowing (the typical leak).
- `f32::NAN` from any caller path.
- A direct `±f32::INFINITY` construction from FFI/WASM/Rust callers.

A finite `0.0` component is intentionally allowed — a zero query vector is a legitimate edge case that the downstream `norm > 0.0` guard in `normalize_in_place` already handles (it skips normalization and returns the zero vector unchanged).

## Environment

- Commit: 5d5dc30
- Rust MSRV in repo: 1.88.0
- Found by: scheduled bug hunt, following up on the BUG-287 / BUG-315 / BUG-322 / BUG-324 / BUG-326 / BUG-328 / BUG-330 chain (query-side sibling of BUG-330's write-side fix in `collect_vector_value`)

davidkelley:
## Summary

Closes #340.

`IndexReader::plan_vectors` in `searchlite-core/src/api/reader.rs` was accepting the caller-provided `VectorQuery::vector` verbatim and passing it through `normalize_in_place` (for the Cosine metric) and into `metric_similarity` / HNSW `search` without validating that each component was finite. An over-range JSON number (e.g. `1e40`) deserializes to a finite `f64` and narrows to `f32` via the default `Deserialize for f32` impl, saturating to `±f32::INFINITY`. Direct Rust/FFI/WASM callers can also construct a `VectorQuery` with `f32::NAN` or `±f32::INFINITY` components.

For the L2 metric the raw `±INF` persists into `hit.vector_score`, which then breaks JSON serialization (`NaN` / `Infinity` are not valid JSON numbers). For Cosine, `normalize_in_place` computes `norm = INF.sqrt() = INF`, then `v /= INF` yields `0.0` for finite components and `NaN` for the offending one; the downstream `dot.is_nan()` guard in `metric_similarity` silently coerces the result to `0.0`, hiding malformed inputs rather than surfacing them.

This is the read/query-side sibling of BUG-330 (PR #331), which added the same-class guard in `collect_vector_value` at insert time. The two guards together keep the contract symmetric — invalid inputs are rejected at the write and read boundaries rather than silently corrupting downstream similarity computations, HNSW traversal, or response serialization.

## Changes

- `searchlite-core/src/api/reader.rs`: validate each component of `vector_query.vector` for finitude inside `plan_vectors`, immediately after the dimension check and before the Cosine `normalize_in_place` call / storage on `VectorClausePlan`. On a non-finite component the function `bail!`s with a message naming the field and the offending index, mirroring the surrounding `bail!` pattern already applied to `alpha` and `boost`.

The check covers all three leak classes:
- finite `f64` magnitude > `f32::MAX` saturating to `±f32::INFINITY` on JSON → `f32` narrowing (the typical leak),
- `f32::NAN` from any caller,
- direct `±f32::INFINITY` construction from FFI / WASM / Rust callers.

A finite `0.0` component is intentionally allowed — a zero query vector is a legitimate edge case that the downstream `norm > 0.0` guard in `normalize_in_place` already handles (skips normalization and returns the zero vector unchanged).

## Tests

Regression tests added in `searchlite-core/tests/vector_search.rs`:

- `vector_query_with_positive_inf_component_is_rejected` — Cosine metric, `QueryNode::Vector` path, `f32::INFINITY` in the first component.
- `vector_query_with_negative_inf_component_is_rejected_for_l2` — L2 metric, `VectorQuerySpec::Structured` hybrid path, `f32::NEG_INFINITY` component.
- `vector_query_with_nan_component_is_rejected` — explicit `f32::NAN` case (not reachable from JSON but reachable from FFI / WASM / Rust callers).
- `vector_query_with_finite_boundary_components_is_accepted` — locks in the boundary so `f32::MAX` / `f32::MIN` pass the guard unchanged, preventing over-rejection of legitimate inputs.

## Validation

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p searchlite-core --lib --all-targets -- -D warnings` — clean
- `cargo clippy -p searchlite-core --features vectors --all-targets -- -D warnings` — clean
- `cargo test -p searchlite-core --features vectors --test vector_search` — 14 / 14 (including 4 new)
- `cargo test -p searchlite-core --features vectors --lib` — 459 / 459

### DIFF
--- a/searchlite-core/src/api/reader.rs
+++ b/searchlite-core/src/api/reader.rs
@@ -487,6 +487,20 @@ impl IndexReader {
           vector_query.vector.len()
         );
       }
+      // Reject non-finite query vector components before they reach
+      // `normalize_in_place` (which yields NaN for cosine when any component
+      // overflows to ±INF) or `metric_similarity` (where L2 surfaces the raw
+      // ±INF into `hit.vector_score` and fails JSON serialization). Mirrors
+      // the write-side guard in `collect_vector_value` (BUG-330).
+      for (idx, value) in vector_query.vector.iter().enumerate() {
+        if !value.is_finite() {
+          bail!(
+            "vector query for field `{}` contains non-finite component at index {}",
+            vector_query.field,
+            idx
+          );
+        }
+      }
       let mut query_vec = vector_query.vector.clone();
       if matches!(
         schema_field.metric,
--- a/searchlite-core/tests/vector_search.rs
+++ b/searchlite-core/tests/vector_search.rs
@@ -691,3 +691,162 @@ fn rejects_global_cap_below_clause_count() {
     "expected validation error, got {err}"
   );
 }
+
+#[test]
+fn vector_query_with_positive_inf_component_is_rejected() {
+  let dir = tempdir().unwrap();
+  let schema = schema();
+  IndexBuilder::create(dir.path(), schema.clone(), opts(dir.path())).unwrap();
+  let idx = Index::open(opts(dir.path())).unwrap();
+  add_docs(
+    &idx,
+    &[Document {
+      fields: [
+        ("_id".into(), serde_json::json!("doc-1")),
+        ("body".into(), serde_json::json!("rust search")),
+        ("embedding".into(), serde_json::json!([1.0, 0.0])),
+      ]
+      .into_iter()
+      .collect(),
+    }],
+  );
+  let reader = idx.reader().unwrap();
+  let req = SearchRequest {
+    query: Query::Node(QueryNode::Vector(VectorQuery {
+      field: "embedding".into(),
+      vector: vec![f32::INFINITY, 0.0],
+      k: Some(3),
+      alpha: Some(0.0),
+      ef_search: None,
+      candidate_size: Some(3),
+      boost: None,
+    })),
+    vector_query: None,
+    vector_filter: None,
+    ..base_request(Query::String("".into()), 3)
+  };
+  let err = reader.search(&req).unwrap_err().to_string();
+  assert!(
+    err.contains("non-finite component") && err.contains("embedding"),
+    "expected non-finite component error, got {err}"
+  );
+}
+
+#[test]
+fn vector_query_with_negative_inf_component_is_rejected_for_l2() {
+  let dir = tempdir().unwrap();
+  let schema = schema_l2();
+  IndexBuilder::create(dir.path(), schema.clone(), opts(dir.path())).unwrap();
+  let idx = Index::open(opts(dir.path())).unwrap();
+  add_docs(
+    &idx,
+    &[Document {
+      fields: [
+        ("_id".into(), serde_json::json!("doc-1")),
+        ("body".into(), serde_json::json!("rust vector")),
+        ("embedding".into(), serde_json::json!([0.0, 0.0])),
+      ]
+      .into_iter()
+      .collect(),
+    }],
+  );
+  let reader = idx.reader().unwrap();
+  let req = SearchRequest {
+    vector_query: Some(VectorQuerySpec::Structured(VectorQuery {
+      field: "embedding".into(),
+      vector: vec![f32::NEG_INFINITY, 0.0],
+      k: Some(3),
+      alpha: Some(0.5),
+      ef_search: None,
+      candidate_size: Some(3),
+      boost: None,
+    })),
+    vector_filter: None,
+    ..base_request(Query::String("rust".into()), 3)
+  };
+  let err = reader.search(&req).unwrap_err().to_string();
+  assert!(
+    err.contains("non-finite component") && err.contains("embedding"),
+    "expected non-finite component error, got {err}"
+  );
+}
+
+#[test]
+fn vector_query_with_nan_component_is_rejected() {
+  let dir = tempdir().unwrap();
+  let schema = schema();
+  IndexBuilder::create(dir.path(), schema.clone(), opts(dir.path())).unwrap();
+  let idx = Index::open(opts(dir.path())).unwrap();
+  add_docs(
+    &idx,
+    &[Document {
+      fields: [
+        ("_id".into(), serde_json::json!("doc-1")),
+        ("body".into(), serde_json::json!("rust search")),
+        ("embedding".into(), serde_json::json!([1.0, 0.0])),
+      ]
+      .into_iter()
+      .collect(),
+    }],
+  );
+  let reader = idx.reader().unwrap();
+  let req = SearchRequest {
+    query: Query::Node(QueryNode::Vector(VectorQuery {
+      field: "embedding".into(),
+      vector: vec![0.0, f32::NAN],
+      k: Some(3),
+      alpha: Some(0.0),
+      ef_search: None,
+      candidate_size: Some(3),
+      boost: None,
+    })),
+    vector_query: None,
+    vector_filter: None,
+    ..base_request(Query::String("".into()), 3)
+  };
+  let err = reader.search(&req).unwrap_err().to_string();
+  assert!(
+    err.contains("non-finite component") && err.contains("embedding"),
+    "expected non-finite component error, got {err}"
+  );
+}
+
+#[test]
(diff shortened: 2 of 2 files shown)


davidkelley:
@codex review

---

PR #343, opened 2026-04-19
fix(sort): reject non-finite search_after _score after f64→f32 cast (BUG-342)

### LINKED ISSUE #342: fix(sort): search_after _score value f64→f32 cast overflow silently corrupts pagination
## Summary

`SortPlan::values_from_json` in `searchlite-core/src/query/sort.rs` decodes the `_score` value from a user-supplied `search_after` token by reading the JSON number as `f64` via `serde_json::Number::as_f64()` and unconditionally casting to `f32` via `f as f32` — without validating that the resulting `f32` is finite. A finite JSON `f64` whose magnitude exceeds `f32::MAX` (~3.4e38) saturates to `±f32::INFINITY` under Rust's `as` cast, and the poisoned `SortValue::Score` is threaded into the cursor's `SortKey` and compared against document scores during pagination. This is the same class of bug as BUG-330 (`collect_vector_value` silent f64→f32 cast) and BUG-336 (scoring-path f64→f32 cast), but in the pagination path instead of indexing or scoring.

The `search_after` token is round-tripped by clients verbatim and is therefore under user control on the decode side, identical to the `script_score` `params` and `bucket_script` `params` maps — both of which `bail!` on non-finite input (`script.rs:177-178`, `aggs/mod.rs:3657-3658`). The `_score` branch has no such guard.

## Affected code

- `searchlite-core/src/query/sort.rs:260-264` — the `SortField::Score` branch of `values_from_json`:
  ```rust
  (SortField::Score, Value::Number(n)) => {
    let f = n
      .as_f64()
      .ok_or_else(|| anyhow::anyhow!("search_after sort value must be number"))?;
    SortValue::Score(f as f32)    // ← no finitude check on the cast result
  }
  ```
  The `as f32` narrowing cast at line 264 saturates silently, and the non-finite `f32` value is stored directly in the `SortValue::Score` variant that feeds `SortKey::compare` via `f32::total_cmp` downstream.

## Reproduction / evidence

A client submits a `search_after` token whose first element is a valid JSON number larger than `f32::MAX`:

```json
{
  "sort": ["_score"],
  "search_after": [1e40, "some-doc-id", 0]
}
```

Evaluation path:

1. `decode_search_after_token` (`searchlite-core/src/api/pagination.rs:228`) calls `sort_plan.values_from_json(&token[..sort_plan.len()])`.
2. `values_from_json` reads `1e40` as `f64` (finite — `f64::MAX` is ~1.8e308).
3. `1e40_f64 as f32` saturates to `f32::INFINITY`.
4. `SortValue::Score(f32::INFINITY)` is stored in the returned `Vec<SortValue>`.
5. `decode_search_after_token` wraps the resulting `SortKey` in `CursorState { key, returned: 0 }`.
6. Subsequent pagination uses `SortKey::cmp` which calls `compare_f32(a, b, order)` → `a.total_cmp(&b)`. Under `total_cmp`, `f32::INFINITY` sorts strictly above every finite `f32` (and above `NaN` in the positive case), so:
   - **Descending sort (`_score` desc, the common case):** the cursor treats `+INF` as "already past every document", and every subsequent page is empty (`.cmp()` → `Ordering::Less` for every candidate). Pagination silently ends early.
   - **Ascending sort:** the cursor treats `+INF` as "before every document", and every document re-appears in the next page. Pagination loops with duplicated results.
7. The symmetric negative case (`-1e40` → `f32::NEG_INFINITY`) inverts the two outcomes.

Via a non-standard deserializer, `f64::INFINITY` / `f64::NAN` in a JSON number would hit the same cast — `NAN as f32 = NAN`, and `total_cmp` assigns `NaN` a defined position in the total order, corrupting ordering but not panicking. The default `serde_json` parser rejects those tokens, so the typical trigger is the finite-f64 overflow above.

Compare with the guarded sibling paths:

- `script_score` `params` (`searchlite-core/src/query/script.rs:177-178`): `bail!("script_score param `{name}` must be finite")`.
- `bucket_script` `params` (`searchlite-core/src/query/aggs/mod.rs:3657-3658`): rejects non-finite at the tokenizer site.
- `collect_vector_value` (BUG-330, PR #331, `searchlite-core/src/index/segment.rs`): `bail!`s on non-finite `f32` immediately after the cast.
- `CompiledFunction::evaluate` / `CompiledScript::evaluate` / `apply_rank_modifier` (BUG-336, PR #337): clamps to `f32::MAX`/`f32::MIN` before the narrowing cast.

The `SortField::Score` decode path has neither guard.

## Expected behavior

A `search_after` token whose `_score` component does not fit in a finite `f32` should cause `values_from_json` to fail with a clear error, matching the existing input-validation pattern in the same function (`bail!` for wrong-type sort values at lines 280, 282, 284, 286) and the `script_score` / `bucket_script` param checks. The `_score` sort value is a response-side artifact — clients are expected to echo back the `f32` value received in `next_search_after`, which is always finite (a non-finite hit score is already rejected by `evaluate_compiled_score` / the BUG-315 guard on `combine_function_scores`). A user-supplied out-of-range value is an API misuse and should surface as an error rather than silently corrupt pagination.

## Actual behavior

`values_from_json` returns `Ok([SortValue::Score(f32::INFINITY), ...])`. The non-finite value is threaded through the cursor into every subsequent `SortKey::cmp` call, silently truncating or duplicating pagination results depending on sort order, with no diagnostic for the caller.

Because `f32::total_cmp` provides a total order over all `f32` values (including `NaN`), no panic or serialization error is produced — the corruption manifests only as wrong pagination output, which is harder to notice than an outright failure.

## Root cause (hypothesis)

The three other `search_after` decode branches (`SortField::Keyword`, `SortField::I64`, `SortField::F64`) either store their value verbatim (keyword, f64) or validate via `as_i64` (i64), which checks the representable range. Only `SortField::Score` narrows via an `as` cast, and the cast was written before the recent BUG-330 / BUG-336 finitude-guard sweep. The pattern of "check finitude immediately after `as f32`" is now established at every other write-side cast site in the crate, but was not back-ported to this decode path.

## Suggested fix

Validate the `f32` form immediately after the cast and `bail!` on non-finite, mirroring the surrounding error pattern:

```rust
(SortField::Score, Value::Number(n)) => {
  let f = n
    .as_f64()
    .ok_or_else(|| anyhow::anyhow!("search_after sort value must be number"))?;
  let score = f as f32;
  if !score.is_finite() {
    bail!("search_after _score value must fit in a finite f32 (received {f})");
  }
  SortValue::Score(score)
}
```

This rejects all three leak classes:

- finite `f64` magnitude > `f32::MAX` saturating to `±f32::INFINITY` (the typical trigger);
- `f64::INFINITY` / `f64::NEG_INFINITY` (only reachable through a non-strict deserializer, but `is_finite()` covers it for free);
- `f64::NAN` (same).

A finite `f64` underflowing to `0.0` is intentionally allowed — `0.0` is finite and a legitimate sort key component.

The three sibling decode branches (`Keyword`, `I64`, `F64`) already handle their invariants: `Keyword` is string-typed so no numeric overflow exists; `I64` uses `as_i64` which rejects out-of-range numbers; `F64` stores the value as `f64` with no narrowing cast. Only the `Score` branch needs the new guard.

## Environment

- Commit: 1e8bae5
- Rust MSRV in repo: 1.88.0
- Found by: scheduled bug hunt, following up on the BUG-330 / BUG-336 f64→f32 narrowing-cast chain

davidkelley:
## Summary

Closes #342.

`SortPlan::values_from_json` in `searchlite-core/src/query/sort.rs` was decoding the `_score` value from a user-supplied `search_after` token by reading the JSON number as `f64` via `serde_json::Number::as_f64()` and unconditionally casting to `f32`. A finite JSON `f64` whose magnitude exceeds `f32::MAX` (~3.4e38) saturates to `±f32::INFINITY` under Rust's `as` cast, and the non-finite `SortValue::Score` was threaded into the cursor's `SortKey`, corrupting pagination via `f32::total_cmp`:

- **Descending sort (the common `_score` default):** `+INF` sorts strictly above every finite candidate, so the cursor treats every subsequent document as "already seen" and pagination silently ends early with empty pages.
- **Ascending sort:** `+INF` sorts above everything, so the cursor re-returns the same documents on each page.
- **`-1e40`:** symmetric `-INF` case that inverts the two outcomes.

The three sibling branches (`Keyword`, `I64`, `F64`) were already safe — `as_i64` rejects out-of-range numbers, `F64` stores the value as `f64` without a narrowing cast, and `Keyword` is string-typed. Only the `Score` branch was missing a guard because the `as f32` narrowing cast predated the BUG-330 / BUG-336 finitude-guard sweep.

This is the same class of bug as BUG-330 (`collect_vector_value` silent f64→f32 cast, PR #331) and BUG-336 (scoring-path f64→f32 cast, PR #337), but on the pagination decode path instead of indexing or scoring.

## Changes

- `searchlite-core/src/query/sort.rs`: narrow the f64 to f32 explicitly, then `bail!` if the result is non-finite, mirroring the existing `bail!` branches in the same function for wrong-type sort values. Error message echoes the offending input.

Policy choice: reject with error rather than clamp. The `search_after` token is a response-side artifact — clients are expected to echo back the `f32` value received in `next_search_after`, which is always finite (non-finite hit scores are already dropped by `evaluate_compiled_score` / the BUG-315 guard on `combine_function_scores`). A user-supplied out-of-range value is an API misuse and should surface as a diagnostic, matching the `script_score` / `bucket_script` params pattern (`script.rs:177-178`, `aggs/mod.rs:3657-3658`).

## Tests

Added regression tests in `searchlite-core/src/query/sort.rs::tests`:

- `values_from_json_rejects_score_overflow_to_positive_inf` — `1e40` is a valid finite JSON number (well within `f64::MAX`) but larger than `f32::MAX`; the cast saturates to `f32::INFINITY` and decode must `bail!`.
- `values_from_json_rejects_score_overflow_to_negative_inf` — symmetric `-1e40` / `f32::NEG_INFINITY` case.
- `values_from_json_accepts_score_at_f32_max` — boundary check: `f32::MAX as f64` round-trips through the cast unchanged and must continue to be accepted to avoid over-rejecting legitimate large scores.
- `values_from_json_accepts_small_finite_score` — sanity test for a typical finite score (`0.5`).

## Validation

- `cargo fmt --all` — clean
- `cargo clippy -p searchlite-core --lib --all-targets -- -D warnings` — clean
- `cargo test -p searchlite-core --lib` — 451 tests pass (includes 4 new `values_from_json_*` cases)

### DIFF
--- a/searchlite-core/src/query/sort.rs
+++ b/searchlite-core/src/query/sort.rs
@@ -261,7 +261,16 @@ impl SortPlan {
           let f = n
             .as_f64()
             .ok_or_else(|| anyhow::anyhow!("search_after sort value must be number"))?;
-          SortValue::Score(f as f32)
+          // Narrow from f64 to f32 explicitly so a finite JSON number larger
+          // than f32::MAX (which Rust's `as` cast saturates to ±f32::INFINITY)
+          // is rejected here rather than poisoning the sort key and silently
+          // corrupting pagination via `total_cmp`. Mirrors the write-side
+          // guards in BUG-330 (collect_vector_value) and BUG-336 (scoring).
+          let score = f as f32;
+          if !score.is_finite() {
+            bail!("search_after _score value must fit in a finite f32 (received {f})");
+          }
+          SortValue::Score(score)
         }
         (SortField::Keyword(_), Value::String(s)) => SortValue::Str(s.clone()),
         (SortField::I64(_), Value::Number(n)) => {
@@ -487,6 +496,76 @@ mod tests {
     };
     assert!(higher_asc > key_asc);
     assert!(higher_desc < key_desc);
+  }
+
+  fn score_sort_plan() -> SortPlan {
+    let schema = Schema {
+      doc_id_field: crate::index::manifest::default_doc_id_field(),
+      analyzers: Vec::new(),
+      text_fields: Vec::new(),
+      keyword_fields: Vec::new(),
+      numeric_fields: Vec::new(),
+      nested_fields: Vec::new(),
+      #[cfg(feature = "vectors")]
+      vector_fields: Vec::new(),
+    };
+    SortPlan::from_request(&schema, &[]).unwrap()
+  }
+
+  #[test]
+  fn values_from_json_rejects_score_overflow_to_positive_inf() {
+    // `1e40` is a valid finite JSON number (well within f64::MAX) but larger
+    // than f32::MAX, so `f as f32` saturates to `f32::INFINITY`. Feeding that
+    // poisoned value into the sort key via `search_after` silently corrupts
+    // pagination comparisons under `total_cmp`; the decode must bail.
+    let plan = score_sort_plan();
+    let raw = vec![serde_json::json!(1.0e40_f64)];
+    let err = plan.values_from_json(&raw).unwrap_err();
+    assert!(
+      err.to_string().contains("finite f32"),
+      "unexpected error: {err}"
+    );
+  }
+
+  #[test]
+  fn values_from_json_rejects_score_overflow_to_negative_inf() {
+    let plan = score_sort_plan();
+    let raw = vec![serde_json::json!(-1.0e40_f64)];
+    let err = plan.values_from_json(&raw).unwrap_err();
+    assert!(
+      err.to_string().contains("finite f32"),
+      "unexpected error: {err}"
+    );
+  }
+
+  #[test]
+  fn values_from_json_accepts_score_at_f32_max() {
+    // Boundary: `f32::MAX as f64` round-trips through the cast unchanged and
+    // must continue to be accepted so legitimate large-but-finite scores are
+    // not over-rejected.
+    let plan = score_sort_plan();
+    let raw = vec![serde_json::json!(f32::MAX as f64)];
+    let values = plan.values_from_json(&raw).unwrap();
+    assert_eq!(values.len(), 1);
+    match &values[0] {
+      SortValue::Score(v) => {
+        assert!(v.is_finite());
+        assert_eq!(*v, f32::MAX);
+      }
+      other => panic!("expected SortValue::Score, got {other:?}"),
+    }
+  }
+
+  #[test]
+  fn values_from_json_accepts_small_finite_score() {
+    let plan = score_sort_plan();
+    let raw = vec![serde_json::json!(0.5_f64)];
+    let values = plan.values_from_json(&raw).unwrap();
+    assert_eq!(values.len(), 1);
+    match &values[0] {
+      SortValue::Score(v) => assert_eq!(*v, 0.5_f32),
+      other => panic!("expected SortValue::Score, got {other:?}"),
+    }
   }
 
   #[test]


davidkelley:
@codex review

PR #348, opened 2026-04-19
fix(aggs): reject non-finite values in parse_interval_seconds (BUG-344)

### LINKED ISSUE #344: fix(aggs): parse_interval_seconds does not reject non-finite results, silently corrupting date_histogram interval and offset
## Summary

`parse_interval_seconds` in `searchlite-core/src/query/aggs/mod.rs` parses a user-supplied duration string (e.g. `"1h"`, `"30m"`) into an `f64` number of seconds but does not verify the result is finite. A numeric prefix whose decimal representation exceeds ~308 digits parses to `f64::INFINITY` via Rust's `f64::from_str` (which returns `Ok(f64::INFINITY)` for out-of-range values rather than `Err`). The infinity propagates through the `* mult` scaling, the `* 1_000.0` millisecond conversion, and the `as i64` saturating cast, silently producing a `DateInterval::Fixed(i64::MAX)` step or an `i64::MAX` offset that corrupts the date_histogram bucketing.

This is the same class of bug that BUG-338 fixed for `parse_date` (which now gates its `f64` fallback with `.filter(|f| f.is_finite())`), but `parse_interval_seconds` was not updated with the same guard.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:4350` — `let value: f64 = rest[..idx].parse().ok()?;` does not check `value.is_finite()`.
- `searchlite-core/src/query/aggs/mod.rs:4361` — `let magnitude = value * mult;` does not check `magnitude.is_finite()` before returning `Some(...)`.
- `searchlite-core/src/query/aggs/mod.rs:1434-1435` — caller converts `parse_interval_seconds` result to `i64` via `millis as i64`, which saturates `f64::INFINITY` to `i64::MAX`.
- `searchlite-core/src/query/aggs/mod.rs:1440-1441` — same pattern for `offset`.

## Reproduction / evidence

A user sends a `date_histogram` aggregation with a `fixed_interval` or `offset` whose numeric prefix is long enough to overflow `f64`:

```json
{
  "aggs": {
    "by_day": {
      "date_histogram": {
        "field": "timestamp",
        "fixed_interval": "999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999h"
      }
    }
  }
}
```

Evaluation path:

1. `parse_interval_seconds` extracts the 310-digit numeric prefix and calls `.parse::<f64>()`, which returns `Ok(f64::INFINITY)`.
2. `value * mult` = `f64::INFINITY * 3600.0` = `f64::INFINITY`.
3. `parse_interval_seconds` returns `Some(f64::INFINITY)`.
4. `DateHistogramCollector::new` computes `millis = f64::INFINITY * 1000.0 = f64::INFINITY`.
5. `DateInterval::Fixed(f64::INFINITY as i64)` saturates to `DateInterval::Fixed(i64::MAX)` (Rust's `as` cast saturates since 1.45).
6. In `bucket_start`, `value.saturating_sub(offset).div_euclid(i64::MAX)` = `0` for any reasonable timestamp → every document lands in a single bucket with key `offset`.
7. In the fill loop, `next_bucket_start` calls `current.checked_add(i64::MAX)` which overflows → returns `None` → fill loop terminates immediately.

**Result:** the aggregation silently returns a single bucket whose key does not correspond to any real date, with no error diagnostic. The user receives nonsensical results.

The symmetric `offset` path is worse: `offset = i64::MAX` causes `value.checked_sub(i64::MAX)` to return `None` for every document, so all documents are silently dropped from the aggregation.

Compare with the guarded sibling:

- `parse_date` (`searchlite-core/src/query/aggs/mod.rs:4320`): `.filter(|f| f.is_finite())` — added by BUG-338 to reject non-finite numeric fallback values.

## Expected behavior

`parse_interval_seconds` should reject non-finite intermediate and final values by returning `None`, mirroring the guard in `parse_date`. A `fixed_interval` or `offset` string that overflows `f64` should surface as a parse failure (the aggregation would use the default interval or offset of 0) rather than silently corrupting bucket keys.

## Actual behavior

`parse_interval_seconds` returns `Some(f64::INFINITY)`. The non-finite value saturates to `i64::MAX` at the `as i64` cast site and silently produces a degenerate aggregation — either all documents in one bucket (for `fixed_interval`) or all documents dropped (for `offset`) — with no diagnostic for the caller.

## Root cause (hypothesis)

BUG-338 added a finitude guard to `parse_date`'s `f64` fallback path because `f64::from_str` accepts `"NaN"`, `"inf"`, and `"infinity"` as valid float literals. `parse_interval_seconds` has the same `f64::from_str` call site (line 4350) but was not included in the BUG-338 fix, likely because the trigger for `parse_interval_seconds` is a different input pattern (a very long digit sequence rather than a literal `"inf"` string) and was not noticed during the sweep. The function's digit-and-dot-only input filter prevents literal `"inf"` from reaching the parser, but does not prevent a long all-digit string from parsing to infinity.

## Suggested fix

Add a finitude filter after the parse and/or after the multiplication:

```rust
let value: f64 = rest[..idx].parse().ok()?;
if !value.is_finite() {
    return None;
}
// ...
let magnitude = value * mult;
if !magnitude.is_finite() {
    return None;
}
```

The double check covers the (unlikely but possible) case where `value` is finite but `value * mult` overflows to infinity.

## Environment

- Commit: 1e8bae5
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Fixes #344.

`parse_interval_seconds` parses the numeric prefix of a duration string (e.g. `"1h"`, `"30m"`) via `f64::from_str`, which returns `Ok(f64::INFINITY)` — not `Err` — for decimal magnitudes exceeding `f64::MAX` (~1.8e308). The digit-and-dot prefix filter prevents literal `"inf"`/`"NaN"` from reaching the parser, but a sufficiently long all-digit prefix still parses to infinity.

Callers in `DateHistogramCollector::new` cast the result to `i64` via `as i64`, which saturates `f64::INFINITY` to `i64::MAX`, silently producing:
- `DateInterval::Fixed(i64::MAX)` — every document collapses into a single bucket at the offset (`div_euclid(i64::MAX) == 0`), and the fill loop exits immediately on `checked_add` overflow.
- `offset == i64::MAX` — every document is dropped from the aggregation (`value.checked_sub(i64::MAX)` returns `None`).

`validate_date_histogram_config` already guarded `fixed_interval` via `secs.is_finite() && secs > 0.0 && (secs * 1000.0) as i64 >= 1`, but the `offset` branch only checks `parse_interval_seconds(offset).is_none()` — so an overflow previously returned `Some(INFINITY)` and bypassed validation entirely.

## Fix

Reject non-finite results inside `parse_interval_seconds` at two sites, mirroring the guard added to `parse_date` for BUG-338:
1. After `f64::from_str` — catches long digit prefixes that parse directly to infinity.
2. After the unit multiplication — catches finite values close to `f64::MAX` that overflow when scaled by a unit multiplier (e.g. `604_800.0` for weeks).

With this guard, overflow surfaces as `None` at the call site, so both the `fixed_interval` and `offset` validators produce the expected `InvalidConfig` error instead of silently corrupting bucket membership.

## Test plan

- [x] New unit tests `parse_interval_seconds_rejects_overflowing_magnitude` and `parse_interval_seconds_rejects_overflow_after_multiplier` cover the post-parse and post-multiplier overflow paths.
- [x] New integration test `date_histogram_rejects_overflowing_interval_strings` in `aggregation_bounds.rs` verifies that a `date_histogram` request with an overflowing `fixed_interval` or `offset` surfaces the expected `InvalidConfig` error at the planner level.
- [x] All existing `searchlite-core` lib tests and `aggregation_bounds` tests pass.
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo clippy -p searchlite-core --all-targets --no-deps -- -D warnings` clean.

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -4348,6 +4348,23 @@ pub(crate) fn parse_interval_seconds(spec: &str) -> Option<f64> {
     return None;
   }
   let value: f64 = rest[..idx].parse().ok()?;
+  // Rust's `f64::from_str` returns `Ok(f64::INFINITY)` for numeric
+  // strings whose decimal magnitude exceeds the `f64` range (~1e308),
+  // rather than `Err`. The digit-and-dot prefix filter above prevents
+  // literal `"inf"`/`"NaN"` strings from reaching the parser, but a
+  // sufficiently long all-digit prefix (e.g. 310+ digits) still parses
+  // to infinity. Downstream call sites cast the result to `i64` via
+  // `as i64`, which saturates `f64::INFINITY` to `i64::MAX` — silently
+  // producing `DateInterval::Fixed(i64::MAX)` (all documents collapse
+  // into a single bucket) or an `i64::MAX` offset (every document is
+  // dropped from the aggregation), with no diagnostic for the caller.
+  // Reject non-finite parse results so overflowed intervals/offsets
+  // surface as a parse failure in the caller, mirroring the
+  // `.filter(|f| f.is_finite())` guard added to `parse_date` for
+  // BUG-338.
+  if !value.is_finite() {
+    return None;
+  }
   let suffix = &rest[idx..];
   let mult = match suffix {
     "" | "s" => 1.0,
@@ -4359,6 +4376,12 @@ pub(crate) fn parse_interval_seconds(spec: &str) -> Option<f64> {
     _ => return None,
   };
   let magnitude = value * mult;
+  // Also reject non-finite scaled magnitudes: a finite `value` close to
+  // `f64::MAX` can still overflow to infinity after multiplication by
+  // the unit multiplier (e.g. `604_800.0` for weeks).
+  if !magnitude.is_finite() {
+    return None;
+  }
   Some(if negative { -magnitude } else { magnitude })
 }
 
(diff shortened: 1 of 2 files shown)


davidkelley:
@codex review

PR #349, opened 2026-04-19
fix(aggs): bucket_script div guard rejects only exact zero (BUG-346)

### LINKED ISSUE #346: fix(aggs): bucket_script eval_rpn division guard rejects small but valid divisors, silently dropping bucket values
## Summary

`eval_rpn` in the `bucket_script` pipeline aggregation uses an epsilon guard (`b.abs() < 1e-12`) for division that silently rejects small but legitimate non-zero divisors, returning `None` (null) instead of the correct finite result. The equivalent division guard in `script_score` (`script.rs`) correctly checks for exact zero (`b == 0.0`) and relies on the post-division `!val.is_finite()` check to catch actual overflow.

## Affected code

- `searchlite-core/src/query/aggs/mod.rs:3684-3689` — `eval_rpn` division branch uses `b.abs() < 1e-12` as the guard
- `searchlite-core/src/query/script.rs:119` — `CompiledScript::evaluate` division branch correctly uses `b == 0.0`

## Reproduction / evidence

```rust
// bucket_script eval_rpn (aggs/mod.rs:3684-3689):
'/' => {
    if b.abs() < 1e-12 {   // <-- rejects any |b| < 1e-12
        return None;
    }
    a / b
}

// script_score (script.rs:119):
if b == 0.0 {               // <-- rejects only exact zero
    return None;
}
let val = a / b;
if !val.is_finite() {        // <-- catches overflow to infinity
    return None;
}
```

Concrete input that triggers the wrong behavior:

```
eval_bucket_script("a / b", {"a": 1.0, "b": 1e-13})
```

- **Expected:** `Some(1e13)` — `1.0 / 1e-13 = 1e13`, a perfectly valid finite f64.
- **Actual:** `None` — the epsilon guard sees `(1e-13).abs() < 1e-12` as true and short-circuits.

An even more damaging case: `eval_bucket_script("a / b", {"a": 1e-20, "b": 1e-13})` should produce `Some(1e-7)` (an ordinary small number), but returns `None`.

## Expected behavior

Division by a non-zero finite divisor should produce the quotient, and only return `None` when:
1. The divisor is exactly `0.0` (or `-0.0`), OR
2. The result is non-finite (caught by the existing `!result.is_finite()` check on line 3692).

This matches the `script_score` behavior and preserves valid arithmetic for small divisors.

## Actual behavior

Any `bucket_script` expression containing a division where the divisor's absolute value is less than `1e-12` silently returns `None`, discarding the bucket's computed value. In a pipeline aggregation context, this surfaces as a missing (null) metric for the affected bucket, with no error or warning.

## Root cause (hypothesis)

The `1e-12` epsilon guard was likely added as a "safe" alternative to exact-zero checking before the non-finite result guard (line 3692) was introduced in commit `68cabb2` (BUG-287). With the finite check now in place, the epsilon guard is both redundant (the finite check already rejects `a / 0.0 = ±Inf`) and incorrect (it rejects valid divisions by small non-zero values). The `script.rs` evaluator was written or updated to use exact-zero checking, but the `eval_rpn` counterpart was not aligned.

## Suggested fix

Replace the epsilon comparison with an exact-zero check, matching `script.rs`:

```rust
'/' => {
    if b == 0.0 {
        return None;
    }
    a / b
}
```

The post-operation `!result.is_finite()` guard on line 3692 already handles the overflow case.

## Environment

- Commit: d795057
- Rust MSRV in repo: 1.88.0

davidkelley:
Closes #346

## Summary

`eval_rpn`'s division branch in `bucket_script` used `b.abs() < 1e-12` as its divisor guard, which silently rejected **valid** small-but-non-zero divisors. The sibling evaluator in `query/script.rs` already does the correct thing: reject only exact zero and rely on the post-operation `!result.is_finite()` check to catch overflow.

### Before (buggy)

```rust
'/' => {
    if b.abs() < 1e-12 {   // rejects any |b| < 1e-12, including valid divisors
        return None;
    }
    a / b
}
```

`eval_bucket_script("a / b", {a: 1.0, b: 1e-13})` returned `None` even though the quotient `1e13` is a perfectly finite `f64`. In a pipeline-aggregation context this surfaced as a silently-missing metric on the affected bucket, with no error.

### After (fixed)

```rust
'/' => {
    if b == 0.0 {
        return None;
    }
    a / b
}
```

Overflow to infinity (e.g. `f64::MAX / f64::MIN_POSITIVE`) is still caught by the existing `!result.is_finite()` guard immediately below the match.

## Changes

- `searchlite-core/src/query/aggs/mod.rs` — replace epsilon divisor guard with `b == 0.0`, mirroring `script.rs`.
- Replace the `bucket_script_rejects_near_zero_division` test (which codified the wrong behavior) with:
  - `bucket_script_rejects_exact_zero_division` — `b = 0.0` still returns None.
  - `bucket_script_rejects_negative_zero_division` — `b = -0.0` still returns None.
  - `bucket_script_accepts_small_non_zero_divisor` — `1.0 / 1e-13 ≈ 1e13` now succeeds.
  - `bucket_script_accepts_tiny_numerator_and_denominator` — `1e-20 / 1e-13 ≈ 1e-7` now succeeds.
  - `bucket_script_rejects_overflow_to_infinity` — overflow still rejected by the existing finite guard.

## Test plan

- [x] `cargo test -p searchlite-core --lib bucket_script` — 28 passed, 0 failed
- [x] `cargo test --workspace --all-features` — all green
- [x] `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean

### DIFF
--- a/searchlite-core/src/query/aggs/mod.rs
+++ b/searchlite-core/src/query/aggs/mod.rs
@@ -3682,7 +3682,11 @@ fn eval_rpn(tokens: Vec<ScriptToken>, vars: &BTreeMap<String, f64>) -> Option<f6
           '-' => a - b,
           '*' => a * b,
           '/' => {
-            if b.abs() < 1e-12 {
+            // Only reject exact zero divisors. Small-but-valid non-zero divisors
+            // (e.g. `1e-13`) produce legitimate finite quotients and must be
+            // preserved. Overflow to infinity is handled by the post-operation
+            // `!result.is_finite()` guard below. Mirrors script.rs (BUG-346).
+            if b == 0.0 {
               return None;
             }
             a / b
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

davidkelley:
@codex review

PR #361, opened 2026-04-20
fix(score): use doc_count for BM25 N so it matches df after deletions (BUG-360)

davidkelley:
## Summary
- Replace `seg.live_docs()` with `seg.meta.doc_count` at the two BM25 scoring call sites in `searchlite-core/src/api/reader.rs` (main search path and rescore path) so N stays consistent with df (`postings.len()` includes deleted documents).
- Deleted documents are still excluded at evaluation time via `seg.is_deleted(...)`, matching the Lucene/Elasticsearch approach and the pattern already used in `examples/pruning.rs`.
- Add a regression test showing that deleting documents that do not share a term's posting list leaves the BM25 score of matching documents unchanged (without the fix, IDF gets clamped to 1.0, dropping the score from 1.762 to 1.0).

## Why
With the old code, a term with df > live_docs would produce `ln(negative)` → NaN, silently clamped to 1.0 by `.max(0.0)`. This distorted IDF whenever segments had deletions, and made cross-segment ranking inconsistent for the same term.

Fixes BUG-360.

## Test plan
- [x] `cargo test --package searchlite-core --lib` (472 tests pass, including new regression)
- [x] `cargo test --workspace --lib`
- [x] Verified new test fails before the fix (`before=1.76214 after=1`) and passes after

### DIFF
--- a/searchlite-core/src/api/reader.rs
+++ b/searchlite-core/src/api/reader.rs
@@ -1726,7 +1726,9 @@ impl IndexReader {
     }
     let explanations: RefCell<HashMap<DocId, HitExplanation>> = RefCell::new(HashMap::new());
 
-    let docs = seg.live_docs() as f32;
+    // BM25 N must use total doc_count (including deleted documents) so it matches
+    // the df basis (postings.len() includes deleted documents). See BUG-360.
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

PR #367, opened 2026-04-20
fix(query): allow infinite WAND upper bounds through pivot finder (BUG-366)

davidkelley:
## Summary

Fixes BUG-366: the WAND pivot-finding loop in `wand_loop` silently skipped terms whose global upper bound was non-finite, guarding accumulation with `bound.is_finite()`. When a query weight is large enough that `bm25 * weight` overflows `f32::MAX`, that term's `ub` becomes `+inf`. A positive-infinity bound is loose but valid — it should immediately satisfy any finite pivot threshold and set the pivot so the term's documents are evaluated. Skipping it leaves the accumulator below the heap threshold, so no pivot is found, the WAND loop exits early, and the overflowing term's documents are silently dropped from the result set.

The fix replaces the `is_finite()` guard with a NaN-only skip: NaN poisons the accumulator and can never compare `>=` to a finite threshold, so it must still be rejected, but infinities now flow through and set the pivot correctly on the first infinite bound encountered.

## Changes

- `searchlite-core/src/query/wand.rs:763-775` — replace `bound.is_finite()` guard with `bound.is_nan()` skip.
- Added regression test `wand_does_not_skip_term_with_infinite_upper_bound` that seeds WAND with a finite-score term and a term whose weight overflows ub to `+inf`; asserts the overflow doc is returned in top-k.

## Test plan

- [x] New regression test fails on `main` (\`doc 100 (overflow term) was dropped from top-k: [1, 2]\`) and passes with the fix.
- [x] `cargo test -p searchlite-core --lib query::wand` — all 9 tests pass.
- [x] `cargo test --workspace --lib` — 475 + 19 + 49 tests pass across the workspace.

### DIFF
--- a/searchlite-core/src/query/wand.rs
+++ b/searchlite-core/src/query/wand.rs
@@ -760,12 +760,19 @@ fn wand_loop<F: FnMut(DocId, f32) -> bool, C: DocCollector + ?Sized>(
         term.upper_bound()
       };
 
-      if bound.is_finite() {
-        acc += bound;
-        if acc >= pivot_threshold {
-          pivot_idx = Some(i);
-          break;
-        }
+      // Skip NaN bounds only: NaN arithmetic poisons the accumulator and
+      // `NaN >= threshold` is always false, so the term could never trigger
+      // the pivot. Positive infinity, on the other hand, is a valid (if
+      // loose) upper bound — it should immediately satisfy any finite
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

davidkelley:
@codex review

PR #368, opened 2026-04-20
fix(query): allow infinite upper bounds in WAND pivot scan (BUG-366)

davidkelley:
## Summary
- Replace `is_finite()` guard in `wand_loop` pivot-scan with `is_nan()` so positive-infinity upper bounds are no longer silently skipped.
- A +inf upper bound is a valid (if loose) pruning bound and trivially satisfies any finite `pivot_threshold`; skipping it was causing the WAND loop to exit early and drop matching documents when an earlier campaign of non-finite hardening introduced the guard.
- NaN bounds are still skipped: `NaN + x = NaN` would poison the accumulator and `NaN >= threshold` is always false.

## Details
The pivot-finding scan accumulates per-term upper bounds until the running sum reaches `pivot_threshold`. When a term's upper bound overflowed to `f32::INFINITY` (e.g. from `score_tf` computing `bm25 * weight` with a very large `weight`), the `is_finite()` check elided both the `acc += bound` and the threshold test. If the remaining finite bounds couldn't reach the threshold on their own, `pivot_idx` stayed `None` and `wand_loop` broke out — so documents the infinite-bound term would have matched were never scored.

Fixes BUG-366.

## Test plan
- [x] New unit tests in `searchlite-core/src/query/wand.rs` reproduce the bug and verify the fix on both the WAND and BMW execution paths:
  - `wand_pivot_scan_accepts_infinite_upper_bound` — proves doc 100 (infinite-bound term) displaces doc 1 in top-1 WAND results.
  - `bmw_pivot_scan_accepts_infinite_block_upper_bound` — same invariant under `ExecutionStrategy::Bmw`.
  - `wand_pivot_scan_still_skips_nan_bound_semantics` — belt-and-braces multi-doc case.
- [x] `cargo test -p searchlite-core` passes (479 lib tests + integration tests).
- [x] `cargo clippy -p searchlite-core --lib --tests` clean.
- [x] `cargo fmt --check` clean.
- [x] Confirmed all three new tests fail on `main` without the fix (`is_finite()` guard → doc 100 missed).

### DIFF
--- a/searchlite-core/src/query/wand.rs
+++ b/searchlite-core/src/query/wand.rs
@@ -760,12 +760,21 @@ fn wand_loop<F: FnMut(DocId, f32) -> bool, C: DocCollector + ?Sized>(
         term.upper_bound()
       };
 
-      if bound.is_finite() {
-        acc += bound;
-        if acc >= pivot_threshold {
-          pivot_idx = Some(i);
-          break;
-        }
+      // Skip NaN bounds only — NaN would poison the accumulator (NaN + x = NaN)
+      // and `NaN >= threshold` is always false under IEEE-754, so it could not
+      // satisfy the threshold anyway. Positive-infinity bounds must be allowed
+      // to flow through: an infinite upper bound is a valid (if loose) bound
+      // meaning "this term cannot be pruned", and it trivially satisfies any
+      // finite pivot_threshold. Skipping it would prevent the pivot from being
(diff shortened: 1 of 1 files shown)


davidkelley:
Closing as a duplicate of #367, which was merged with the same fix (replace the `is_finite()` pivot-scan guard with `is_nan()`). Picked up as a race between parallel scheduled bug-bash runs.

PR #372, opened 2026-04-20
fix(score): reject non-finite Constant score from boost-product overflow (BUG-370)

### LINKED ISSUE #370: fix(score): evaluate_compiled_score Constant variant does not reject non-finite score from accumulated boost overflow
## Summary

The `CompiledScoreNode::Constant` variant in `evaluate_compiled_score` returns
`Some(*score)` without a finitude check. Every other scoring variant (Sum,
DisMax, FunctionScore, RankFeature, ScriptScore) guards its output with
`is_finite()`, but Constant does not. When accumulated boosts overflow `f32` to
`+inf` during query planning, the non-finite score leaks through the WAND loop
into `Hit.score`, causing `serde_json` serialization to fail (infinity is not
valid JSON) and returning a 500 error to the caller.

## Affected code

- `searchlite-core/src/api/scoring.rs:337-343` — `CompiledScoreNode::Constant`
  returns `Some(*score)` without checking `score.is_finite()`.
- `searchlite-core/src/query/planner.rs:774-775` — `ScoreNode::Constant` stores
  `boost * node_boost` without checking the product for finitude. Each factor is
  individually validated by `validate_boost` (finite, non-negative), but their
  product can overflow `f32::MAX` to `+inf`.
- `searchlite-core/src/api/scoring.rs:144-147` — `compile_score_node` passes the
  score through to `CompiledScoreNode::Constant` without validation.
- `searchlite-core/src/api/scoring.rs:322-323` — comment claims "child scores are
  guarded by their respective nodes", but this invariant is violated by the
  unguarded Constant variant.

## Reproduction / evidence

A query with nested boosts whose product exceeds `f32::MAX` (~3.4e38):

```json
{
  "query": {
    "bool": {
      "boost": 1e38,
      "must": [{
        "constant_score": {
          "filter": {"match_all": {}},
          "boost": 1e38
        }
      }]
    }
  }
}
```

Trace through the planner:

1. Root `build_node` is called with `boost = 1.0`.
2. Bool node: `child_boost = boost * node_boost = 1.0 * 1e38 = 1e38` (finite).
3. ConstantScore child: `score = boost * node_boost = 1e38 * 1e38 = 1e76` — this
   exceeds `f32::MAX` and saturates to `+inf`.
4. Because the Bool has a single `must` clause, the `ScoreNode::Constant` is used
   directly as the Bool's score node (not wrapped in `Sum`), so neither the Sum
   nor DisMax finitude guards apply.
5. `compile_score_node` stores `score = +inf` without validation.
6. At query time, `evaluate_compiled_score` reaches the `Constant` arm at
   `scoring.rs:337` and returns `Some(+inf)`.
7. The score flows through the `score_adjust` closure in `reader.rs:1836` →
   into the WAND heap → into `Hit.score`.
8. `serde_json::to_string(&result)` fails: "float must be finite".

The same issue applies to any nesting depth that causes the accumulated boost to
overflow — e.g., two levels of `bool` with `boost: 1e20` each.

## Expected behavior

The `Constant` scoring node should reject non-finite scores (returning `None`),
consistent with every other variant of `evaluate_compiled_score`. Documents that
would receive a non-finite constant score should be dropped from the result set
rather than causing a serialization failure.

## Actual behavior

`Some(+inf)` leaks into the result, causing `serde_json` serialization to fail
and the HTTP endpoint to return a 500 error. In the library API, the caller
receives a `Hit` with `score: f32::INFINITY`, which cannot be serialized to JSON.

## Root cause (hypothesis)

Two gaps that reinforce each other:

1. **Missing guard in `CompiledScoreNode::Constant`**: the `evaluate` match arm
   returns `Some(*score)` unconditionally. All other arms check `is_finite()` on
   their output.

2. **Unvalidated accumulated boost product**: `build_node` multiplies the
   inherited `boost` by `node_boost` at each nesting level without checking
   whether the product overflows `f32`. The product is stored in
   `ScoreNode::Constant { score }` and later in `CompiledScoreNode::Constant`.

Either fix alone would close the hole: rejecting non-finite values in the
Constant evaluator (defensive) or validating the boost product at planning time
(proactive). Both are likely warranted.

## Suggested fix

Add a finitude guard in the `Constant` evaluator, matching the pattern used by
every other variant:

```rust
CompiledScoreNode::Constant { score, matcher } => {
  if evaluator.matches_subquery(matcher, doc_id) {
    if !score.is_finite() {
      return None;
    }
    Some(*score)
  } else {
    Some(0.0)
  }
}
```

Optionally, also validate the boost product at planning time in `build_node` for
`ConstantScore` (and potentially other nodes that store `boost * node_boost`):

```rust
let score = boost * node_boost;
if !score.is_finite() {
  bail!("accumulated boost overflows f32 in constant_score");
}
```

## Environment

- Commit: 1c43b2c
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

Closes #370.

The `Constant` arm of `evaluate_compiled_score` in `searchlite-core/src/api/scoring.rs` returned `Some(*score)` unconditionally. Every other scoring variant — `Sum`, `DisMax`, `FunctionScore`, `RankFeature`, `ScriptScore` — already guards its output with `is_finite()`, but `Constant` did not. When `boost * node_boost` accumulated across nested scopes overflowed `f32::MAX` to `+INFINITY`, the non-finite score leaked into the WAND sort-key heap (silently corrupting ordering) and into `Hit.score`, where `serde_json` rejects it as invalid JSON and the HTTP endpoint returns a 500.

Each factor in the boost product is already validated to be finite and non-negative by `validate_boost` in the planner, but the **product** can still saturate: e.g. a `Bool` with `boost: 1e38` wrapping a `ConstantScore` with `boost: 1e38` yields `1e76`, which exceeds `f32::MAX ≈ 3.4e38`. The Bool's single-`must` shortcut lifts the `ScoreNode::Constant` directly into the score tree without wrapping it in a `Sum`, so the existing `Sum` / `DisMax` finitude guards never see the non-finite value.

## Changes

- `searchlite-core/src/api/scoring.rs` — add an `is_finite()` guard to the `CompiledScoreNode::Constant` arm of `evaluate_compiled_score`, returning `None` (skip document) when the stored score is non-finite. Mirrors the policy already used by every other arm of this match.

- `searchlite-core/tests/function_score.rs` — two new regression tests:
  - `constant_score_rejects_document_when_boost_product_overflows_to_infinity` — a `Bool { boost: 1e38, must: [ConstantScore { boost: 1e38, ... }] }` whose product overflows to `+INFINITY`. With the fix, the hit is dropped; confirmed to fail on `main` with `got 2 hits with scores [inf, inf]`.
  - `constant_score_keeps_document_when_boost_product_stays_finite` — boundary check that `1e10 * 1e10 = 1e20` (comfortably within `f32::MAX`) is still accepted and scored correctly, guarding against over-rejection.

## Validation

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p searchlite-core --all-targets -- -D warnings` — clean
- `cargo clippy -p searchlite-core --features vectors --all-targets -- -D warnings` — clean
- `cargo test -p searchlite-core` — all 18 suites green, including 52/52 in `function_score` (2 new regression tests)
- `cargo test -p searchlite-core --features vectors --test function_score` — 52/52 green

## Test plan

- [x] Regression test reproduces the bug on `main` and passes with the fix
- [x] Boundary test confirms finite-but-large scores are preserved
- [x] Existing `function_score` suite unaffected (50 prior tests still green)
- [x] Full `searchlite-core` suite green with and without the `vectors` feature
- [x] Clippy clean on both default and `vectors` feature sets

### DIFF
--- a/searchlite-core/src/api/scoring.rs
+++ b/searchlite-core/src/api/scoring.rs
@@ -336,6 +336,16 @@ pub(crate) fn evaluate_compiled_score(
     }
     CompiledScoreNode::Constant { score, matcher } => {
       if evaluator.matches_subquery(matcher, doc_id) {
+        // The stored score is `boost * node_boost` accumulated from every
+        // enclosing scope during planning. Each factor is validated to be
+        // finite, but the product can still overflow `f32::MAX` to
+        // `+INFINITY`. Every other variant of this match rejects non-finite
+        // outputs; Constant must too so the non-finite score does not leak
+        // into the WAND heap (where it corrupts ordering) or into `Hit.score`
+        // (where `serde_json` rejects it as invalid JSON, returning 500).
+        if !score.is_finite() {
+          return None;
+        }
         Some(*score)
       } else {
         Some(0.0)
--- a/searchlite-core/tests/function_score.rs
+++ b/searchlite-core/tests/function_score.rs
@@ -1844,3 +1844,81 @@ fn script_score_large_negative_literal_overflow_clamps_to_f32_min() {
     );
   }
 }
+
+// Regression: the `Constant` arm of `evaluate_compiled_score` previously
+// returned `Some(*score)` unconditionally. Every other variant (Sum, DisMax,
+// FunctionScore, RankFeature, ScriptScore) rejects non-finite results, but
+// Constant did not. When `boost * node_boost` accumulated across nested
+// scopes overflowed `f32::MAX` to `+INFINITY`, the non-finite score leaked
+// into the WAND heap (corrupting ordering) and into `Hit.score`, where
+// `serde_json` rejects it as invalid JSON and the HTTP endpoint returns 500.
+// See BUG-370.
+#[test]
+fn constant_score_rejects_document_when_boost_product_overflows_to_infinity() {
+  let reader = setup_reader();
+  // `Bool` with `boost = 1e38` containing a single `ConstantScore` with
+  // `boost = 1e38`. Both factors are individually finite and pass
+  // `validate_boost`, but their product `1e38 * 1e38 = 1e76` overflows
+  // `f32::MAX ≈ 3.4e38` and saturates to `+INFINITY`. With the guard in
+  // place, the Constant evaluator returns `None` and the hit is dropped.
+  let req = base_request(QueryNode::Bool {
+    must: vec![QueryNode::ConstantScore {
+      filter: Filter::KeywordEq {
+        field: "lang".into(),
+        value: "en".into(),
+      },
+      boost: Some(1e38),
+    }],
+    should: Vec::new(),
+    must_not: Vec::new(),
+    filter: Vec::new(),
+    minimum_should_match: None,
+    boost: Some(1e38),
+  });
+  let resp = reader.search(&req).unwrap();
+  assert!(
+    resp.hits.is_empty(),
+    "expected no hits when Constant score overflows to infinity, got {} hits with scores {:?}",
+    resp.hits.len(),
+    resp.hits.iter().map(|h| h.score).collect::<Vec<_>>()
+  );
+}
+
+#[test]
+fn constant_score_keeps_document_when_boost_product_stays_finite() {
+  let reader = setup_reader();
+  // Boundary check: the new finitude guard must not over-reject legitimate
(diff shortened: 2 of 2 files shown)


davidkelley:
@codex review

PR #380, opened 2026-04-20
fix(score): reject non-finite decay factor (BUG-379)

### LINKED ISSUE #379: fix(score): decay factor range check does not reject NaN, silently producing wrong scores
## Summary

The `compile_functions` range check for the decay `decay` factor uses ordered comparisons (`decay <= 0.0 || decay > 1.0`) that are always `false` for `NaN` under IEEE 754 semantics, allowing `NaN` to be stored as the decay factor. This produces silently wrong scores: the Linear decay function collapses every document to score 0.0, while Exp/Gauss drop all documents not at the exact origin.

## Affected code

- `searchlite-core/src/query/score_functions.rs:96-98` — the range check that NaN bypasses
- `searchlite-core/src/query/score_functions.rs:200-206` — `decay_value()` where the NaN propagates into scoring

## Reproduction / evidence

Send a search request via the WASM binding (`serde_wasm_bindgen` passes JS `NaN` into `f64`) with:

```json
{
  "query": {
    "type": "function_score",
    "query": { "type": "match_all" },
    "functions": [{
      "type": "decay",
      "field": "popularity",
      "origin": 0.0,
      "scale": 10.0,
      "decay": NaN,
      "function": "linear"
    }],
    "boost_mode": "replace"
  }
}
```

**Trace through the code:**

1. `compile_functions` at line 96: `let decay = decay.unwrap_or(0.5);` → `f64::NAN`
2. Line 97: `if NaN <= 0.0 || NaN > 1.0` → `false || false` → `false` — check passes
3. NaN is stored in `CompiledFunction::Decay { decay: NaN, .. }`
4. At evaluation time, `decay_value(NaN, norm, &Linear)`:
   - `((1.0 - norm) * (1.0 - NaN) + NaN).max(0.0)`
   - `= ((1.0 - norm) * NaN + NaN).max(0.0)`
   - `= NaN.max(0.0)`
   - `= 0.0` (Rust's `f64::max` follows IEEE 754-2008 maxNum: returns the non-NaN argument)
5. Every document receives score 0.0 — silently wrong but finite, so the `is_finite()` guard at line 190 does not catch it.

For Exp/Gauss, `NaN.powf(norm)` returns NaN for all `norm > 0`, which is caught by the finitude guard — but the documents are silently dropped instead of the request being rejected at the boundary.

## Expected behavior

`compile_functions` should reject a non-finite `decay` factor at the boundary, returning an error like `"decay factor must be finite"`, consistent with the `is_finite()` guards already applied to `origin` (line 81), `scale` (line 84), and `offset` (line 88) by BUG-373.

## Actual behavior

- **Linear**: All documents receive score 0.0 regardless of distance from origin. The user sees results ordered by a flat score — the decay has no effect, but no error is reported.
- **Exp/Gauss**: Documents at the exact origin get score 1.0 (`NaN.powf(0.0) = 1.0` per IEEE 754); all other documents are silently dropped (NaN score → `None`). The user sees only origin-matching documents with no explanation.

## Root cause (hypothesis)

The BUG-373 fix added `is_finite()` guards for `origin` and `offset` to match the existing `scale` guard, but the `decay` factor was overlooked because it already had a range check. The range check was written assuming well-formed numeric input, but under IEEE 754 semantics, NaN is not ordered — `NaN <= 0.0` and `NaN > 1.0` are both `false`, so the bail condition is never triggered.

## Suggested fix

Add an `is_finite()` check before the range check, matching the pattern used for origin/scale/offset:

```rust
let decay = decay.unwrap_or(0.5);
if !decay.is_finite() {
    bail!("decay factor must be finite");
}
if decay <= 0.0 || decay > 1.0 {
    bail!("decay factor must be in the range (0, 1]");
}
```

Alternatively, rewrite the range check to use an affirmative finite-range test: `if !(decay > 0.0 && decay <= 1.0)` — this rejects NaN, negative zero, infinities, and out-of-range values in one condition.

## Environment

- Commit: f1169a2
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- Reject non-finite (`NaN`/`+Inf`/`-Inf`) decay factors at `compile_functions` in `searchlite-core/src/query/score_functions.rs`, matching the finitude guards already applied to `origin`/`scale`/`offset`.
- Previously, the existing `decay <= 0.0 || decay > 1.0` range check used ordered comparisons that are always `false` for `NaN` under IEEE-754, so a NaN factor (e.g. a JS `NaN` passed through `serde_wasm_bindgen`) slipped through. For `Linear`, every document silently collapsed to score `0.0`; for `Exp`/`Gauss`, documents away from the origin were silently dropped by the `is_finite()` evaluation guard.
- Adds regression tests covering `NaN`, `+Inf`, and `-Inf` decay factors.

Closes #379.

## Test plan

- [x] `cargo test --package searchlite-core --test function_score decay`
- [x] `cargo test --package searchlite-core --lib`
- [x] `cargo clippy --package searchlite-core --all-targets -- -D warnings`

### DIFF
--- a/searchlite-core/src/api/scoring.rs
+++ b/searchlite-core/src/api/scoring.rs
@@ -336,6 +336,16 @@ pub(crate) fn evaluate_compiled_score(
     }
     CompiledScoreNode::Constant { score, matcher } => {
       if evaluator.matches_subquery(matcher, doc_id) {
+        // The stored score is `boost * node_boost` accumulated from every
+        // enclosing scope during planning. Each factor is validated to be
+        // finite, but the product can still overflow `f32::MAX` to
+        // `+INFINITY`. Every other variant of this match rejects non-finite
+        // outputs; Constant must too so the non-finite score does not leak
+        // into the WAND heap (where it corrupts ordering) or into `Hit.score`
+        // (where `serde_json` rejects it as invalid JSON, returning 500).
+        if !score.is_finite() {
+          return None;
+        }
         Some(*score)
       } else {
         Some(0.0)
--- a/searchlite-core/src/query/score_functions.rs
+++ b/searchlite-core/src/query/score_functions.rs
@@ -94,6 +94,9 @@ pub(crate) fn compile_functions(
           bail!("decay scale must be > 0");
         }
         let decay = decay.unwrap_or(0.5);
+        if !decay.is_finite() {
+          bail!("decay factor must be finite");
+        }
         if decay <= 0.0 || decay > 1.0 {
           bail!("decay factor must be in the range (0, 1]");
         }
--- a/searchlite-core/tests/function_score.rs
+++ b/searchlite-core/tests/function_score.rs
@@ -1844,3 +1844,81 @@ fn script_score_large_negative_literal_overflow_clamps_to_f32_min() {
     );
   }
 }
+
+// Regression: the `Constant` arm of `evaluate_compiled_score` previously
+// returned `Some(*score)` unconditionally. Every other variant (Sum, DisMax,
+// FunctionScore, RankFeature, ScriptScore) rejects non-finite results, but
+// Constant did not. When `boost * node_boost` accumulated across nested
+// scopes overflowed `f32::MAX` to `+INFINITY`, the non-finite score leaked
+// into the WAND heap (corrupting ordering) and into `Hit.score`, where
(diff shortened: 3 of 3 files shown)


davidkelley:
@codex review

PR #389, opened 2026-04-21
fix(vectors): saturate l2_distance overflow to f32::MAX (BUG-388)

### LINKED ISSUE #388: fix(vectors): l2_distance silently overflows to +inf between two validated L2 vectors, silently dropping docs from results
## Summary

`l2_distance` (`searchlite-core/src/vectors/mod.rs:104-111`) accumulates `(x - y)^2` into `sum` without any overflow guard and returns `sum.sqrt()` unchanged. Even with the per-vector sum-of-squares validation added by BUG-384 (cosine) and BUG-386 (L2, in-flight on PR #387), two L2 vectors that each pass the `sum_sq.is_finite()` boundary can still produce an overflowing pairwise distance:

- BUG-386 guarantees `sum(v_i^2) <= f32::MAX ≈ 3.4e38` for every indexed or queried L2 vector, which bounds `|v_i| <= sqrt(f32::MAX) ≈ 1.84e19`.
- But `|a_i - b_i|` can still reach `2 * 1.84e19 = 3.68e19`, so `(a_i - b_i)^2` can reach `~1.35e39 > f32::MAX` in a **single dimension**.
- `sum += d * d` saturates to `+inf`, `sqrt(+inf) = +inf`, `metric_similarity(L2) = -inf`.
- `vector_scores[(seg, doc)] = -inf * boost` flows into `compute_hybrid_score`, which catches the non-finite `vector_sum` via the BUG-328 guard (`reader.rs:231`) and returns `None` — the caller silently sees a missing doc with no error.

Same silent-drop symptom as BUG-386, on the orthogonal input shape BUG-386 explicitly flagged as a needed follow-up:

> **BUG-386 — Suggested fix (excerpt)**: *"Add a lightweight overflow guard inside `l2_distance` itself as belt-and-braces for any code path that bypasses validation (explanations, HNSW construction-time similarity, future callers)."*

PR #387 addresses the sum-of-squares validation at the write/query boundary but does **not** include this `l2_distance` internal guard; the `vectors/mod.rs` file is not touched by that PR.

## Affected code

- `searchlite-core/src/vectors/mod.rs:104-111` — `l2_distance` accumulates `d * d` without overflow check and returns `+inf` unchanged.
- `searchlite-core/src/vectors/mod.rs:113-126` — `metric_similarity` negates the overflowed distance and returns `-inf`.
- `searchlite-core/src/vectors/hnsw.rs:189, 215, 262` — HNSW entry, candidate-expansion, and `prune_list` similarity all call `metric_similarity` and thread `-inf` through `total_cmp` ordering, distorting neighbor selection at graph-construction time.
- `searchlite-core/src/api/reader.rs:231` — the BUG-328 guard in `compute_hybrid_score` then drops the candidate, producing the silent-empty-hit symptom.

## Reproduction / evidence

Schema with an L2 vector field, `dim: 2`. Assume PR #387 is merged (both vectors below pass BUG-386 at commit time and plan time):

```json
{
  "embedding": { "type": "array", "items": { "type": "number" },
                 "searchlite:vector": { "dim": 2, "metric": "L2" } }
}
```

**Trigger:**

1. Index doc_a: `{"embedding": [1.5e19, 0.0]}`. `sum_sq = (1.5e19)^2 = 2.25e38 < f32::MAX ≈ 3.4e38`, passes BUG-386.
2. Index doc_b: `{"embedding": [-1.5e19, 0.0]}`. Same: `sum_sq = 2.25e38`, passes BUG-386.
3. Query with `{"vector": [1.5e19, 0.0], "metric": "L2", "k": 10}`. Query `sum_sq = 2.25e38`, passes BUG-386.
4. Against doc_b: `l2_distance([1.5e19, 0], [-1.5e19, 0])`:
   - `d[0] = 1.5e19 - (-1.5e19) = 3.0e19` — finite
   - `d[0] * d[0] = 9.0e38` — **overflows `f32::MAX = 3.4e38` to `+inf`**
   - `sum = +inf`, `sqrt(+inf) = +inf`
   - `metric_similarity(L2) = -(+inf) = -inf`
5. `vector_scores[(seg, doc_b)] = -inf * 1.0 = -inf`
6. `compute_hybrid_score` catches `!vector_sum.is_finite()` and returns `None`
7. Caller sees only doc_a in results; doc_b is silently missing with no error

Verified numerically with `rustc`:

```text
sum_sq(a) = 2.25e38 finite=true
sum_sq(b) = 2.25e38 finite=true
f32::MAX = 3.4e38
  d = 3e19, d*d = inf, sum = inf finite=false
l2_distance = inf finite=false
metric_similarity(L2) = -inf
```

## Expected behavior

`l2_distance` — an internal similarity primitive consumed by HNSW construction, HNSW search, explanations, and hybrid scoring — should always return a finite value. When the true distance would exceed the `f32` range, saturate to `f32::MAX` so the doc ranks at the bottom of the distance ordering (matching the `missing_vector_score(L2) = f32::MIN` sentinel used elsewhere for unrepresentable / unavailable distances) rather than silently dropping out of the result set via the downstream `compute_hybrid_score` guard.

This preserves all three invariants callers rely on:
- HNSW `Scored::cmp` / `total_cmp` ordering remains well-defined (no `-inf` ties with finite scores).
- `compute_hybrid_score` never sees a non-finite `vector_sum` from `l2_distance`, so the BUG-328 guard stops being a silent-drop for legitimately-indexed docs.
- Explanations and HNSW graph construction get a finite, deterministically-ordered distance even for extreme pairwise magnitudes.

## Actual behavior

- **Query path**: a doc that is validly indexed is silently dropped from results when the query vector is "far" from it (pairwise squared difference in a single dimension exceeds `f32::MAX`). No error, no hit, no diagnostic — the caller cannot tell it from a true "no such doc" result.
- **HNSW construction path**: `add_vector` → `search_internal` → `similarity` threads `-inf` through `candidates` / `results` heaps, and `prune_list` sorts overflowing pairs with `-inf` at the bottom. The graph's neighbor lists for surrounding docs can be subtly distorted by the overflowed pair.
- **Explanation path**: when `explain: true` is requested and the explanation walks `l2_distance` directly, the non-finite value can leak into the JSON payload (serialization rejects `+inf`/`NaN`, causing HTTP 500).

## Root cause

`l2_distance` was written without an overflow guard because all callers were expected to validate at the boundary. BUG-386 closes the per-vector sum-of-squares gap, but the per-dimension-difference sum-of-squared-differences is not bounded by per-vector sum-of-squares: worst-case `sum((a_i - b_i)^2) ≈ 3 * f32::MAX` even when both `sum(a_i^2)` and `sum(b_i^2)` are `<= f32::MAX`. The internal primitive still needs its own guard.

## Suggested fix

Make `l2_distance` saturate to `f32::MAX` when the accumulated sum overflows. The resulting distance is finite, orderable, and cleanly separates overflowing pairs from legitimate large-but-representable distances:

```rust
// searchlite-core/src/vectors/mod.rs
pub fn l2_distance(a: &[f32], b: &[f32]) -> f32 {
  let mut sum = 0.0f32;
  for (x, y) in a.iter().zip(b.iter()) {
    let d = x - y;
    sum += d * d;
  }
  let dist = sum.sqrt();
  // Saturate to a finite value so callers (HNSW construction/search,
  // compute_hybrid_score, explanations) never observe a non-finite distance
  // from two individually-validated vectors whose pairwise squared diff
  // overflows `f32::MAX` in any dimension. Returning `f32::MAX` keeps the
  // doc ranked at the bottom of the distance order rather than silently
  // dropping it via the BUG-328 hybrid guard. Belt-and-braces for BUG-386.
  if dist.is_finite() {
    dist
  } else {
    f32::MAX
  }
}
```

Regression tests:
- Unit: `l2_distance([1.5e19, 0], [-1.5e19, 0])` must return a finite value.
- Integration: an L2 query vector far from an indexed L2 doc (both individually valid) returns the doc in results with a ranking-consistent score rather than an empty hit set.

## Environment

- Commit: 7e69ddf (origin/main); independent of PR #387 (does not touch `vectors/mod.rs`).
- Rust MSRV in repo: 1.88.0
- Found by: scheduled bug hunt, follow-up to BUG-386's own "belt-and-braces" recommendation.

davidkelley:
## Summary

Fixes BUG-388 (#388). BUG-384 / BUG-386 bound every indexed and queried L2 vector to `sum(v_i^2) <= f32::MAX`, which constrains `|v_i| <= sqrt(f32::MAX) ≈ 1.84e19` per component. But pairwise `|a_i - b_i|` can still reach `3.68e19`, so a single dimension's squared difference `(a_i - b_i)^2` can exceed `f32::MAX` in the `l2_distance` accumulator — even when both vectors pass every validation at the boundary.

Before this fix: `sum` saturated to `+inf`, `sqrt(+inf) = +inf`, `metric_similarity(L2) = -inf` flowed into `compute_hybrid_score`, and the BUG-328 non-finite guard silently dropped the candidate. The caller saw a missing doc from a legitimately-indexed document with no error or diagnostic — the same silent-drop symptom BUG-386 was meant to prevent, on the orthogonal input shape BUG-386's own "Suggested fix" section explicitly flagged as a needed follow-up:

> *"Add a lightweight overflow guard inside `l2_distance` itself as belt-and-braces for any code path that bypasses validation (explanations, HNSW construction-time similarity, future callers)."*

PR #387 addresses the sum-of-squares validation at the write/query boundary but does not touch `vectors/mod.rs`; this PR is orthogonal.

## Reproduction (numerically verified)

```text
a = [1.5e19, 0]      sum(a_i^2) = 2.25e38 < f32::MAX (3.4e38)  — passes BUG-386
b = [-1.5e19, 0]     sum(b_i^2) = 2.25e38 < f32::MAX            — passes BUG-386
(a[0] - b[0])^2      = (3e19)^2 = 9e38 > f32::MAX               — OVERFLOWS to +inf
l2_distance(a, b)    = +inf
metric_similarity(L2) = -inf                                    — silently dropped
```

## Changes

- **`searchlite-core/src/vectors/mod.rs`** (`l2_distance`)
  - Check `dist.is_finite()` after `sum.sqrt()`. Non-finite result → saturate to `f32::MAX` so downstream callers (HNSW construction/search, `compute_hybrid_score`, explanation payload) always observe a finite distance and the doc ranks at the bottom of the L2 order rather than silently disappearing. The saturation value is consistent with the existing `missing_vector_score(L2) = f32::MIN` sentinel.

### Tests

- **`searchlite-core/src/vectors/mod.rs`** (unit)
  - `l2_distance_saturates_to_f32_max_when_squared_diff_overflows` — the exact pairwise-overflow pair from the reproduction above: both vectors pass BUG-386 but produce an overflowing distance. Asserts the distance is finite **and** equal to `f32::MAX`.
  - `metric_similarity_l2_stays_finite_when_l2_distance_overflows` — end-to-end contract: `metric_similarity(L2)` must never return non-finite when its inputs pass validation.
  - `l2_distance_preserves_finite_result_within_range` — boundary test: `sqrt(4e18) = 2e9` must round-trip through the new guard without being truncated, so we don't over-clamp legitimate distances.
  - `l2_distance_returns_exact_value_for_small_vectors` — sanity: the 3-4-5 triangle is unchanged.

- **`searchlite-core/tests/vector_search.rs`** (integration)
  - `l2_search_returns_far_doc_when_pairwise_squared_diff_overflows` — indexes both `near = [1.5e19, 0]` and `far = [-1.5e19, 0]`, queries with `[1.5e19, 0]`, and asserts both are returned with finite scores. Before the fix, `far` was silently dropped by the `compute_hybrid_score` BUG-328 guard.

## Test plan

- [x] `cargo test -p searchlite-core --features vectors --lib vectors::` — 9 tests pass (includes 4 new BUG-388 unit regressions).
- [x] `cargo test -p searchlite-core --features vectors --test vector_search` — 21 tests pass (includes the new BUG-388 integration regression).
- [x] `cargo test -p searchlite-core --features vectors` — full core matrix green: 524 lib tests + all integration suites.
- [x] `cargo fmt --all --check` — clean.
- [x] Reverted the `l2_distance` guard locally and confirmed both new unit regressions (`l2_distance_saturates_to_f32_max_…`, `metric_similarity_l2_stays_finite_…`) **and** the integration regression (`l2_search_returns_far_doc_…`) fail before the fix — the tests genuinely exercise the bug rather than tautologically passing.

Closes #388

### DIFF
--- a/searchlite-core/src/vectors/mod.rs
+++ b/searchlite-core/src/vectors/mod.rs
@@ -107,7 +107,25 @@ pub fn l2_distance(a: &[f32], b: &[f32]) -> f32 {
     let d = x - y;
     sum += d * d;
   }
-  sum.sqrt()
+  let dist = sum.sqrt();
+  // BUG-388: saturate to `f32::MAX` when the accumulated sum overflows.
+  // BUG-384 / BUG-386 bound every indexed and queried vector to
+  // `sum(v_i^2) <= f32::MAX`, which constrains `|v_i| <= sqrt(f32::MAX) ≈
+  // 1.84e19` per component. Pairwise `|a_i - b_i|` can still reach
+  // `3.68e19`, so `(a_i - b_i)^2` can exceed `f32::MAX` in a single
+  // dimension (e.g. `a = [1.5e19, 0]`, `b = [-1.5e19, 0]` gives `(3e19)^2
+  // = 9e38 > f32::MAX`). Without a guard here, `sum` saturates to `+inf`,
+  // `sqrt(+inf) = +inf`, `metric_similarity(L2) = -inf`, and the BUG-328
+  // hybrid-score guard silently drops the doc from results. Returning a
+  // finite sentinel keeps the ordering well-defined (the doc ranks at the
+  // bottom, consistent with `missing_vector_score(L2) = f32::MIN`) and
+  // prevents non-finite scores from leaking into HNSW graph construction,
+  // `compute_hybrid_score`, or the explanation payload.
+  if dist.is_finite() {
+    dist
+  } else {
+    f32::MAX
+  }
 }
 
 pub fn metric_similarity(metric: &VectorMetric, a: &[f32], b: &[f32]) -> f32 {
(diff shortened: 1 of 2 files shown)


davidkelley:
@codex review

---

PR #395, opened 2026-04-22
fix(vectors): saturate vscore * boost overflow to f32 range (BUG-394)

### LINKED ISSUE #394: fix(vectors): per-candidate `vscore *= clause.boost` can overflow to non-finite, silently dropping L2 candidates
## Summary

`collect_vector_maps` (`searchlite-core/src/api/reader.rs:955`) multiplies the per-candidate HNSW similarity by the user-supplied clause boost with a raw `f32 *= f32` step:

```rust
for (doc_id, mut vscore) in candidates.into_iter() {
  ...
  vscore *= clause.boost;
  ...
}
```

Each factor is individually validated finite and non-negative at plan time (`reader.rs:557-560`), but the **product** can still overflow past `f32::MAX` to `±INFINITY`. After BUG-388, `l2_distance` saturates to `f32::MAX` specifically so `metric_similarity(L2)` can return a finite sentinel (`-f32::MAX = f32::MIN`) at the worst pair — but any user-supplied `boost > 1.0` then turns that finite sentinel straight back into `-INF` at the very next step, and the BUG-328 hybrid-score guard silently drops the candidate with no diagnostic.

In other words, BUG-388's saturation contract is undone for every vector query that sets a non-default boost, and the exact silent-drop symptom BUG-388 was written to eliminate is reintroduced.

Same class of bug as BUG-370 / BUG-381: a plan-time finitude guard covers each input, but the arithmetic combination at runtime can still overflow and is not guarded.

## Affected code

- `searchlite-core/src/api/reader.rs:955` — `vscore *= clause.boost;` performs raw `f32` multiplication with no saturation. The resulting non-finite value is inserted into `vector_scores[(seg, doc)]` and carried into `compute_hybrid_score`, where the non-finite guard at `reader.rs:231` drops the candidate unconditionally.
- `searchlite-core/src/vectors/mod.rs:104-129` — `l2_distance` saturates to `f32::MAX` specifically so downstream `metric_similarity(L2)` never returns `-INF`; the guarantee holds only as far as the next multiplication step.

## Reproduction / evidence

Schema with an L2 vector field, `dim: 2`:

```json
{ "embedding": { "type": "array", "items": { "type": "number" },
                 "searchlite:vector": { "dim": 2, "metric": "L2" } } }
```

**Write-side setup (both vectors pass every existing validation):**

```json
{"_id": "near", "embedding": [1.5e19, 0.0]}
{"_id": "far",  "embedding": [-1.5e19, 0.0]}
```

Each vector has `sum(v_i^2) = 2.25e38 < f32::MAX ≈ 3.4e38`, so both pass BUG-330 (per-component finite) and BUG-386/PR#390 (sum-of-squares finite).

**Query with a non-default boost:**

```json
{ "query": { "match_all": {} },
  "vector_query": { "field": "embedding", "vector": [1.5e19, 0.0],
                    "alpha": 0.0, "k": 10, "boost": 2.0 } }
```

Trace:
1. HNSW returns `[(near, 0.0), (far, f32::MIN)]`. The `far` vscore comes from BUG-388's saturation: `l2_distance(far, query)` saturates to `f32::MAX` (the pairwise `d[0]^2 = (3e19)^2 = 9e38 > f32::MAX` overflows the accumulator), and `metric_similarity(L2) = -l2_distance = f32::MIN`.
2. `vscore *= 2.0` for `far`: `f32::MIN * 2.0` overflows to `-INF`.
3. `vector_scores[(seg, far)] = -INF` is inserted into the map.
4. In `compute_hybrid_score`: `vector_sum += -INF → -INF`; with `alpha = 0`, `blended = vec_score = -INF`; `blended_sum = -INF`; `final_score = -INF / 1 = -INF`.
5. The BUG-328 guard (`reader.rs:231`) rejects non-finite `final_score` / `vector_sum` and returns `None`.
6. The `far` candidate is silently dropped from results — no error, no hit, no diagnostic.

This is the exact failure mode BUG-388's regression test (`l2_search_returns_far_doc_when_pairwise_squared_diff_overflows`) defends against. The regression test passes today because it uses `boost: None` (default `1.0`); introducing any `boost > 1.0` reintroduces the drop.

The same arithmetic pattern can also fire for Cosine, but cosine vectors are normalized so `|vscore| <= 1` and `vscore * boost` stays finite unless `boost > f32::MAX` (which is already rejected). L2 is the observable case in practice.

## Expected behavior

Same contract as BUG-388: a non-finite intermediate must not leak into `vector_scores` / `compute_hybrid_score`. Saturate `vscore * boost` to the f32 range so the candidate is returned (ranked at the bottom under L2 "higher is better" semantics) instead of being silently dropped.

## Actual behavior

`vscore *= clause.boost;` performs raw f32 multiplication. For any `(vscore, boost)` pair whose product exceeds `f32::MAX` in magnitude, the result is `±INF`, which then flows into `vector_scores` and is silently filtered out by the BUG-328 hybrid-score guard. The caller gets a result set missing the far doc even though every input was valid.

## Root cause

The clause boost is validated at plan time, and `l2_distance` is saturated (BUG-388), but the *product* of the already-saturated vscore and the validated boost has no guard. The saturation contract is therefore local to `l2_distance` and does not survive the next multiplication step.

## Suggested fix

Compute the product in f64 (which cannot overflow for any finite f32 × f32) and clamp back to the f32 range. This matches the `f64_to_finite_f32` pattern already used by BUG-336 / BUG-342 for the same class of cast-side overflow:

```rust
// searchlite-core/src/api/reader.rs — collect_vector_maps
let scaled = (vscore as f64) * (clause.boost as f64);
vscore = if scaled >= f32::MAX as f64 {
  f32::MAX
} else if scaled <= f32::MIN as f64 {
  f32::MIN
} else {
  scaled as f32
};
```

Or equivalently, reuse the existing helper:

```rust
vscore = f64_to_finite_f32((vscore as f64) * (clause.boost as f64));
```

Either form preserves every legitimate in-range product as-is and keeps the saturated `f32::MIN` sentinel finite under any `boost > 1.0`.

## Environment

- Commit: ded039e (origin/main)
- Rust MSRV in repo: 1.88.0
- Feature flag: `vectors`
- Found by: scheduled bug hunt, follow-up to BUG-388 / BUG-386 on the uncovered per-candidate-boost path.

davidkelley:
## Summary

Fixes BUG-394 (#394). BUG-388 saturates `l2_distance` to `f32::MAX` so `metric_similarity(L2)` can return the finite sentinel `f32::MIN` at the worst pair instead of `-INF`, which the BUG-328 hybrid-score guard would otherwise drop. The saturation is local to `l2_distance`, though — the very next step in `collect_vector_maps` is `vscore *= clause.boost`, and any user-supplied `boost > 1.0` turns the `f32::MIN` sentinel straight back into `-INF` via f32 multiplication overflow. The non-finite `vscore` then flows into `vector_scores`, `compute_hybrid_score` rejects it, and the far doc is silently dropped again — undoing BUG-388's contract for any query with a non-default boost.

Same class of bug as BUG-370 / BUG-381: a plan-time finitude guard covers each input, but the arithmetic combination at runtime can still overflow and is not guarded.

## Changes

- **`searchlite-core/src/api/reader.rs`** (`collect_vector_maps`)
  - Replace raw `vscore *= clause.boost;` with `vscore = f64_to_finite_f32((vscore as f64) * (clause.boost as f64));`. Every finite `f32 × f32` product fits in f64 without overflow; `f64_to_finite_f32` then clamps to `[f32::MIN, f32::MAX]` so overflowing products saturate rather than producing `±INF`. Mirrors the `f64_to_finite_f32` clamping policy already used by BUG-336 / BUG-342 on the scoring-side cast.

### Tests (`searchlite-core/tests/vector_search.rs`)

- **New** `l2_search_with_boost_returns_far_doc_when_pairwise_squared_diff_overflows` — identical to BUG-388's `l2_search_returns_far_doc_when_pairwise_squared_diff_overflows` regression except for `boost: Some(2.0)`. Verified pre-fix: `far` disappears from the result set (`cargo test` fails with `assert!(ids.contains(&\"far\"))`). Post-fix: `far` is returned with a finite vector score, matching BUG-388's contract under any `boost > 1.0`.
- **New** `vector_boost_preserves_finite_product_within_range` — cosine similarity `1.0 × boost 3.0 == 3.0` must survive the f64 clamp unchanged, so the saturation cannot silently perturb legitimate in-range scores.

## Test plan

- [x] `cargo test -p searchlite-core --features vectors --test vector_search` — all 24 vector integration tests pass, including the two new BUG-394 regressions.
- [x] `cargo test -p searchlite-core --features vectors` — full core matrix green (524 lib tests + every integration suite).
- [x] `cargo clippy -p searchlite-core --all-targets -- -D warnings` — clean (no-vectors build path).
- [x] Reverted the src change locally and confirmed the new `l2_search_with_boost_returns_far_doc_...` test fails (the `far` doc is silently dropped, matching #394's reproduction) while every pre-existing test still passes — the regression genuinely exercises the bug rather than tautologically passing.

Closes #394

### DIFF
--- a/searchlite-core/src/api/reader.rs
+++ b/searchlite-core/src/api/reader.rs
@@ -32,6 +32,8 @@ use crate::query::collector::{AggregationSegmentCollector, DocCollector};
 use crate::query::filters::passes_filter;
 use crate::query::planner::{build_query_plan, QueryMatcher, ScorePlan};
 use crate::query::sort::{SortKey, SortPlan};
+#[cfg(feature = "vectors")]
+use crate::query::util::f64_to_finite_f32;
 use crate::query::wand::{
   execute_top_k_with_stats_and_mode_internal, score_tf, QueryStats, ScoreAdjustFn, ScoreMode,
   ScoredTerm,
@@ -952,7 +954,21 @@ impl IndexReader {
               continue;
             }
           }
-          vscore *= clause.boost;
+          // BUG-394: saturate the `vscore * boost` product to the f32 range.
+          // Both factors are validated finite at plan time (per-component
+          // guards and `boost.is_finite()`), but the raw `f32 *= f32` can
+          // still overflow to `±INF` — notably for L2, where BUG-388 saturates
+          // `l2_distance` to `f32::MAX` so `metric_similarity(L2)` can return
+          // `f32::MIN` at the worst pair. Any user-supplied `boost > 1.0`
+          // would then turn that finite sentinel back into `-INF`, which
+          // `compute_hybrid_score`'s BUG-328 guard silently drops — undoing
+          // BUG-388's contract that far docs rank last rather than disappear.
+          // Computing in f64 (which cannot overflow for any finite f32 × f32)
+          // and clamping back to the f32 range keeps every legitimate in-range
+          // product exact while preventing the non-finite leak. Mirrors the
+          // `f64_to_finite_f32` clamping policy introduced for BUG-336 /
+          // BUG-342 on the scoring-side cast.
+          vscore = f64_to_finite_f32((vscore as f64) * (clause.boost as f64));
           let cand = VectorCandidate {
             segment_ord: segment_ord as u32,
             doc_id,
--- a/searchlite-core/tests/vector_search.rs
+++ b/searchlite-core/tests/vector_search.rs
@@ -1152,3 +1152,153 @@ fn l2_search_returns_far_doc_when_pairwise_squared_diff_overflows() {
     );
   }
 }
+
+// BUG-394: BUG-388 saturates `l2_distance` to `f32::MAX` so that
+// `metric_similarity(L2)` returns the finite sentinel `f32::MIN` at the worst
+// pair instead of `-INF`, which the BUG-328 hybrid-score guard would otherwise
+// drop. The saturation is local to `l2_distance`, though — the very next step
+// in `collect_vector_maps` is `vscore *= clause.boost`, and any user-supplied
+// `boost > 1.0` turns the `f32::MIN` sentinel straight back into `-INF` via
+// f32 multiplication overflow. That non-finite vscore then flows into
+// `vector_scores`, `compute_hybrid_score` rejects it, and the `far` doc is
+// silently dropped again — undoing BUG-388's fix for any query with a
+// non-default boost. The regression test below is identical to BUG-388's
+// `l2_search_returns_far_doc_when_pairwise_squared_diff_overflows` except for
+// `boost: Some(2.0)`; before the fix that single change causes `far` to
+// disappear from results, matching the failure mode BUG-388 was written to
+// eliminate.
+#[test]
+fn l2_search_with_boost_returns_far_doc_when_pairwise_squared_diff_overflows() {
+  let dir = tempdir().unwrap();
+  let schema = schema_l2();
+  IndexBuilder::create(dir.path(), schema.clone(), opts(dir.path())).unwrap();
+  let idx = Index::open(opts(dir.path())).unwrap();
+  add_docs(
+    &idx,
+    &[
+      Document {
+        fields: [
+          ("_id".into(), serde_json::json!("near")),
+          ("body".into(), serde_json::json!("rust vector")),
+          ("embedding".into(), serde_json::json!([1.5e19_f64, 0.0_f64])),
+        ]
+        .into_iter()
+        .collect(),
+      },
+      Document {
+        fields: [
+          ("_id".into(), serde_json::json!("far")),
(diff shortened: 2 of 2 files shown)


davidkelley:
Closing as duplicate of #398, which was merged first with the identical source fix (`vscore = f64_to_finite_f32((vscore as f64) * (clause.boost as f64))`) and slightly stronger tests (`.expect()` on `vector_score` plus an end-to-end `hit.score ≈ vector_score` assertion for vector-only queries). Rebasing this branch onto main resolves all conflicts to a zero-diff commit.

PR #400, opened 2026-04-22
fix(query): reject non-finite tie_breaker at plan time (BUG-399)

### LINKED ISSUE #399: fix(query): validate_tie_breaker accepts NaN, silently dropping or wrong-scoring DisMax / MultiMatch hits
## Summary

`validate_tie_breaker` (`searchlite-core/src/query/planner.rs:996-1005`) range-checks `tie_breaker` against `< 0.0` and `> 1.0`, but neither comparison fires for `NaN` under IEEE-754 (every ordered comparison with NaN returns `false`). A NaN `tie_breaker` is therefore accepted at plan time and baked into both `ScoreNode::DisMax.tie_breaker` (used by the regular scoring path in `searchlite-core/src/api/scoring.rs`) and `ScoreExpr::DisMax.tie_breaker` (used by the BM25 fast-path in `searchlite-core/src/query/planner.rs`).

At runtime, `max + tie_breaker * (sum - max)` evaluates to `NaN`. The defensive `is_finite()` guards downstream then either silently drop the document (`scoring.rs:329-331` — `evaluate_compiled_score` returns `None`, the doc is excluded from results) or clamp the score to `0.0` (`planner.rs:205-209` — the doc keeps matching but its DisMax contribution is wrong). The caller sees a missing or mis-ranked hit with no actionable diagnostic.

Same class of bug as BUG-373 / BUG-379: a numeric validator uses ordered comparisons against bounds without an `is_finite()` precondition, so NaN slips through and corrupts downstream scoring. The sibling validator one function up — `validate_boost` (`planner.rs:935-941`) — explicitly guards `!value.is_finite()` for exactly this reason; `validate_tie_breaker` should be symmetric.

## Affected code

- `searchlite-core/src/query/planner.rs:996-1005` — `validate_tie_breaker` checks `< 0.0` / `> 1.0` only; both are `false` for `NaN`, so `Ok(NaN)` is returned.
- `searchlite-core/src/query/planner.rs:444` — `MultiMatch` calls `validate_tie_breaker` on the user-supplied `tie_breaker`.
- `searchlite-core/src/query/planner.rs:534` — `DisMax` calls the same validator on the user-supplied `tie_breaker`.
- `searchlite-core/src/api/scoring.rs:288-336` — runtime DisMax: `max + NaN * (sum - max) == NaN`; the BUG-364 `!result.is_finite()` guard at `:329-331` then returns `None`, dropping the doc.
- `searchlite-core/src/query/planner.rs:172-209` — BM25 fast-path DisMax: same arithmetic, the `if result.is_finite()` guard at `:205-209` clamps to `0.0`, producing a wrong score for the doc instead of a drop.

## Reproduction / evidence

The JSON path naturally rejects `NaN` during deserialisation (it is not valid JSON), but the WASM binding (`searchlite-wasm/src/wasm.rs:952` — `serde_wasm_bindgen::from_value::<SearchRequest>(...)`) accepts JS `NaN` directly — the same path that surfaced BUG-373 and BUG-379. A minimal Rust-level reproduction:

```rust
use searchlite_core::api::types::QueryNode;

let req = base_request(QueryNode::DisMax {
  queries: vec![
    QueryNode::Match { /* hits doc-1, doc-2 */ },
  ],
  tie_breaker: Some(f32::NAN),
  boost: None,
});
let result = reader.search(&req).unwrap();   // succeeds, instead of erroring
assert!(result.hits.is_empty());              // matching docs silently dropped
```

Trace:

1. `validate_tie_breaker(Some(NaN))`:
   - `value = NaN`
   - `NaN < 0.0` → `false`
   - `NaN > 1.0` → `false`
   - returns `Ok(NaN)` — **bug**
2. `ScoreNode::DisMax { children, tie_breaker: NaN }` is materialised at plan time.
3. `evaluate_compiled_score` for `DisMax`:
   - `max + NaN * (sum - max)` evaluates to `NaN`
   - `if !result.is_finite() { return None }` fires at `scoring.rs:329-331`
   - the candidate is silently excluded from the top-k heap
4. The HTTP / WASM caller receives an empty (or short) hit set with no error.

The fast-path `ScoreExpr::DisMax` produces the equivalent failure on the pure-BM25 path, except the non-finite guard there clamps to `0.0` (`planner.rs:205-209`), so the doc keeps matching but its DisMax contribution silently collapses to zero — wrong ranking instead of a drop.

## Expected behaviour

Reject non-finite `tie_breaker` at plan time with an actionable validation error, mirroring the existing `validate_boost` finitude guard at `planner.rs:937` and the `decay` finitude guards merged in BUG-373 / BUG-379. Surfacing a deterministic plan-time error gives the caller something they can act on instead of a silent missing-hits bug at runtime.

## Actual behaviour

- `DisMax` regular-scoring path (`scoring.rs`): every matching candidate is silently dropped, the response is shorter than expected with no diagnostic.
- `DisMax` BM25 fast-path (`planner.rs::ScoreExpr::DisMax`): every matching candidate is kept but scored as `0.0`, producing a wrong ranking with no diagnostic.
- `MultiMatch { tie_breaker: NaN }` reaches the same DisMax-shaped scoring code with the same two failure modes depending on which path is selected.

## Root cause

`validate_tie_breaker` was written when the only known invalid inputs were the explicit `< 0` / `> 1` cases. NaN was not considered, and `NaN < 0.0` / `NaN > 1.0` are both `false` under IEEE-754, so it slips through. The sibling validator `validate_boost` immediately above already encodes the right pattern (`!value.is_finite() || value.is_sign_negative()`), and the BUG-373 / BUG-379 fixes added the same `is_finite()` guard to every other user-supplied f32/f64 in the function-score and decay families.

## Suggested fix

Add an `is_finite()` precondition to `validate_tie_breaker`, before the range checks, mirroring `validate_boost`:

```rust
fn validate_tie_breaker(tie: &Option<f32>) -> Result<f32> {
  let value = tie.unwrap_or(0.0);
  if !value.is_finite() {
    bail!("tie_breaker must be finite");
  }
  if value < 0.0 {
    bail!("tie_breaker must be non-negative");
  }
  if value > 1.0 {
    bail!("tie_breaker must be <= 1.0");
  }
  Ok(value)
}
```

The single `is_finite()` check covers `NaN`, `+Infinity`, and `-Infinity` (the existing `< 0.0` would catch `-Infinity` and the existing `> 1.0` would catch `+Infinity`, but stating it explicitly is symmetric with `validate_boost` and unambiguously rejects `NaN`). Both `MultiMatch` and `DisMax` callers benefit because they share the validator.

## Environment

- Commit: 7662816 (origin/main)
- Rust MSRV in repo: 1.88.0
- Found by: scheduled bug hunt, follow-up to BUG-373 / BUG-379 / BUG-381 on the same NaN-slips-through-ordered-comparison pattern.

davidkelley:
## Summary

Fixes BUG-399 (#399). `validate_tie_breaker` (`searchlite-core/src/query/planner.rs:996-1005`) range-checks `tie_breaker` against `< 0.0` and `> 1.0`, but neither comparison fires for `NaN` under IEEE-754 (every ordered comparison with NaN returns `false`). A NaN `tie_breaker` is therefore accepted at plan time and baked into both `ScoreNode::DisMax.tie_breaker` (regular scoring path in `searchlite-core/src/api/scoring.rs:288-336`) and `ScoreExpr::DisMax.tie_breaker` (BM25 fast-path in `searchlite-core/src/query/planner.rs:172-209`).

At runtime `max + NaN * (sum - max) = NaN`. The downstream finitude guard in `evaluate_compiled_score` (BUG-364) at `scoring.rs:329-331` then returns `None` and the candidate is silently excluded from the top-k heap; the parallel guard on the BM25 fast path (`planner.rs:205-209`) instead clamps to `0.0`, keeping the doc but giving its DisMax contribution a wrong score. Both `MultiMatch` and `DisMax` callers reach the same scoring code via the shared validator and exhibit the same two failure modes.

Same class of bug as BUG-373 / BUG-379: a numeric validator uses ordered comparisons against bounds without an `is_finite()` precondition, so NaN slips through and corrupts downstream scoring. The sibling validator one function up — `validate_boost` (`planner.rs:935-941`) — already encodes the right pattern (`!value.is_finite() || value.is_sign_negative()`); `validate_tie_breaker` is now symmetric.

The JSON path naturally rejects `NaN`/`Infinity` during deserialisation, but `serde_wasm_bindgen` (`searchlite-wasm/src/wasm.rs:952`) accepts JS `NaN` directly — the same path that surfaced BUG-373 and BUG-379.

## Changes

- **`searchlite-core/src/query/planner.rs`** (`validate_tie_breaker`)
  - Add an `is_finite()` precondition before the existing range checks, with a metric-neutral `tie_breaker must be finite` error and a doc-comment explaining the IEEE-754 ordered-comparison gap and the WASM trigger. The single guard covers `NaN`, `+Infinity`, and `-Infinity` (the existing `> 1.0` would also catch `+Infinity` and `< 0.0` would catch `-Infinity`, but stating it explicitly is symmetric with `validate_boost` and unambiguously rejects `NaN`). Both `MultiMatch` and `DisMax` callers benefit because they share the validator.

### Tests (`searchlite-core/src/query/planner.rs::tests`)

- **New** `validate_tie_breaker_rejects_nan` — `validate_tie_breaker(Some(NaN))` must surface the finitude error instead of returning `Ok(NaN)`.
- **New** `validate_tie_breaker_rejects_positive_infinity` / `validate_tie_breaker_rejects_negative_infinity` — symmetric to `validate_boost`'s finitude treatment; the error message must be the unified `tie_breaker must be finite` regardless of which infinity sign was supplied.
- **New** `validate_tie_breaker_accepts_finite_in_range` — `None` (default `0.0`), `0.0`, `0.5`, and `1.0` must still round-trip, so the new guard cannot over-reject legitimate inputs.
- **New** `validate_tie_breaker_still_rejects_out_of_range` — `-0.1` and `1.1` must still produce the existing `non-negative` / `<= 1.0` errors, so the new guard does not displace the existing range checks.
- **New** `build_query_plan_rejects_dis_max_tie_breaker_nan` — `QueryNode::DisMax { tie_breaker: NaN }` must surface a plan-time error instead of silently dropping every matching candidate at runtime via the BUG-364 guard in `evaluate_compiled_score`.
- **New** `build_query_plan_rejects_multi_match_tie_breaker_nan` — `QueryNode::MultiMatch { tie_breaker: NaN }` must surface the same plan-time error; `MultiMatch` reaches the same scoring code via the same validator.

## Test plan

- [x] `cargo test -p searchlite-core --lib query::planner` — all 27 planner tests pass, including the 7 new BUG-399 regressions.
- [x] `cargo test -p searchlite-core --lib` — full lib matrix green (508 tests).
- [x] `cargo test -p searchlite-core` — full core matrix green (508 lib tests + every integration suite).
- [x] Reverted the `validate_tie_breaker` change locally and confirmed all 5 `rejects_*` regressions fail (NaN slips through with no error; ±Infinity is caught by the existing range checks but produces the wrong error text) while the `accepts_finite_in_range` and `still_rejects_out_of_range` controls still pass — the regressions genuinely exercise the bug rather than tautologically passing.

Closes #399

### DIFF
--- a/searchlite-core/src/query/planner.rs
+++ b/searchlite-core/src/query/planner.rs
@@ -995,6 +995,19 @@ fn clamp_expansions(
 
 fn validate_tie_breaker(tie: &Option<f32>) -> Result<f32> {
   let value = tie.unwrap_or(0.0);
+  // Every ordered comparison with `NaN` returns `false` under IEEE-754, so the
+  // `< 0.0` and `> 1.0` range checks below do not fire for `NaN`. Without this
+  // guard a caller supplying `tie_breaker: NaN` (reachable via the WASM
+  // binding, same as BUG-373 / BUG-379) plants `NaN` into
+  // `ScoreNode::DisMax.tie_breaker` and `ScoreExpr::DisMax.tie_breaker`. At
+  // runtime `max + NaN * (sum - max)` evaluates to `NaN`, which either
+  // silently drops every candidate via the `is_finite()` guard in
+  // `evaluate_compiled_score` or clamps their DisMax contribution to `0.0`
+  // via the matching guard on the BM25 fast path. Reject non-finite
+  // `tie_breaker` at the plan boundary instead, mirroring `validate_boost`.
+  if !value.is_finite() {
+    bail!("tie_breaker must be finite");
+  }
   if value < 0.0 {
     bail!("tie_breaker must be non-negative");
   }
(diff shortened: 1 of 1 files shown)


davidkelley:
@codex review

---

davidkelley:
@codex review

PR #402, opened 2026-04-22
fix(score): reject term_weights accumulation overflow (BUG-401)

### LINKED ISSUE #401: fix(score): term_weights accumulation overflows to non-finite when multiple query clauses share the same field:term key
## Summary

`term_weights` accumulation in `IndexReader::search` (and the parallel sites in `rescore_hits` and the suggestion path) sums per-term weights with `entry.1 += term.weight` without a finitude check on the running total. Each individual `term.weight` is validated finite by `combine_boost` (BUG-381 / BUG-396), but when multiple query nodes produce qualified terms that share the same `field:term` key — e.g. two `bool.should` clauses both matching `body:rust` — the weights are summed into a single `ScoredTerm.weight`. Two large-but-finite boosts whose sum exceeds `f32::MAX` (~3.4e38) overflow to `+Inf`, which then propagates through `score_tf` (`base * Inf = Inf`) into the WAND/brute-force scoring pipeline.

Same class of bug as BUG-381 / BUG-394 / BUG-396: each input is individually validated at plan or expansion time, but the runtime arithmetic combination is unguarded.

## Affected code

- `searchlite-core/src/api/reader.rs:1271` — main search path: `entry.1 += term.weight` with no finitude check on the accumulated total.
- `searchlite-core/src/api/reader.rs:2104` — rescore path: identical unguarded accumulation.
- `searchlite-core/src/api/reader.rs:4054` — suggestion/explain path: identical unguarded accumulation.
- `searchlite-core/src/query/wand.rs:313` — `score_tf` returns `base * weight`; with `weight = Inf`, every scored document gets `Inf`.
- `searchlite-core/src/query/wand.rs:121-130` — `upper_bound_tf` computes an `Inf` upper bound, distorting WAND pivot selection.

## Reproduction / evidence

A `bool` query with two `should` clauses matching the same field and term, each carrying a boost whose sum exceeds `f32::MAX`:

```json
{
  "query": {
    "bool": {
      "should": [
        { "match": { "body": { "query": "rust", "boost": 1.8e38 } } },
        { "match": { "body": { "query": "rust", "boost": 1.8e38 } } }
      ]
    }
  }
}
```

Trace:

1. `validate_boost(Some(1.8e38))` → `Ok(1.8e38)` — each boost is finite and non-negative. ✓
2. `combine_boost(1.0, 1.8e38)` → `Ok(1.8e38)` — product with parent boost is finite. ✓
3. Both clauses produce `QualifiedTerm { key: "body:rust", weight: 1.8e38, leaf: 0 }` and `QualifiedTerm { key: "body:rust", weight: 1.8e38, leaf: 1 }`.
4. `term_weights` accumulation: first insert sets weight = 1.8e38; second `+= 1.8e38` yields `3.6e38 > f32::MAX ≈ 3.4e38` → **`+Inf`**.
5. `ScoredTerm { weight: Inf, ... }` is built and passed to the WAND/brute-force executor.
6. `score_tf(tf, df, doc_len, avgdl, docs, k1, b, Inf)` → `bm25_base * Inf = Inf` for every matching document.
7. In the WAND loop, `score_sum = Inf`; `plan.evaluate(buf)` with `ScoreExpr::Sum` clamps the `Inf` to `0.0` (per BUG-364's guard) — producing a **wrong score** instead of the correct positive BM25 value.
8. On the brute-force path without a compound `ScoreExpr`, the raw `Inf` enters the heap, corrupting `heap_threshold` and potentially causing `serde_json::to_vec` to fail on the response (HTTP 500).

The same trigger works with a query-string query containing duplicate terms (`"rust rust"` with a large boost), or with `N` should clauses whose per-clause boost `B` satisfies `N * B > f32::MAX`.

A secondary issue: the `term_weights` accumulation at line 1271 silently drops the `leaf` index of the second (and subsequent) qualified terms sharing the same key — the entry retains the first term's leaf. This means the second clause's scoring leaf never receives any contribution, further distorting the score.

## Expected behavior

Reject the query with an actionable validation error when the accumulated weight for any field:term key overflows `f32` to a non-finite value, mirroring the `combine_boost` guard that rejects non-finite products at plan time (BUG-381) and expansion time (BUG-396). The guard should fire before the weight enters `ScoredTerm`, giving the caller a diagnosable failure mode instead of a silent wrong-score or serialization failure.

## Actual behavior

- **WAND + ScoreExpr::Sum path**: every matching document is scored as `0.0` (the `Sum` finitude guard clamps the `Inf` leaf contribution), producing a wrong ranking where affected documents sort to the bottom.
- **Brute-force path (no compound ScoreExpr)**: the `Inf` score leaks into the heap. `serde_json::to_vec` fails on `+Inf`, returning HTTP 500. If the response is consumed by the library API without serialization, the caller sees `Inf` scores in the result set.
- **WAND upper-bound path**: `upper_bound_tf(..., Inf) = Inf`, so the affected term's upper bound is infinite. While this does not cause incorrect pruning (an infinite bound is an overestimate, which is safe for WAND), it defeats block-max skip optimizations for that term, degrading performance.

## Root cause

`term_weights` accumulation was introduced to merge duplicate keys across query groups into a single `ScoredTerm`. The summation assumes that the individual weights (each validated finite by `combine_boost`) cannot produce a non-finite total. This holds for typical queries, but a `bool` query with `N ≥ 2` should clauses matching the same field:term key allows each clause to contribute its own weight to the sum, and the sum of `N` finite f32 values can overflow `f32::MAX` when individual values are large.

## Suggested fix

Add a finitude check after each accumulation, at all three sites:

```rust
entry.1 += term.weight;
if !entry.1.is_finite() {
    bail!(
        "accumulated boost for term `{}` overflows f32; \
         reduce per-clause boost values",
        term.key
    );
}
```

This mirrors the `combine_boost` pattern: reject at the earliest point where the non-finite value is produced, before it can leak into the scoring pipeline.

## Environment

- Commit: 7662816 (origin/main)
- Rust MSRV in repo: 1.88.0

davidkelley:
## Summary

- Individual term boosts are validated finite by `combine_boost` at plan and expansion time (BUG-381 / BUG-396), but when multiple query clauses resolve to the same `field:term` key the per-term weights are summed into a single `ScoredTerm.weight`. The sum of N finite f32 values can still overflow `f32::MAX` (~3.4e38) to `+Inf`, which then propagates through `score_tf` into the WAND / brute-force scoring pipeline — silently clamped to `0.0` under BUG-364's `Sum` guard or leaked as an `Inf` score that trips `serde_json::to_vec` (HTTP 500).
- Add a finitude check at both production accumulation sites (`IndexReader::search` and `IndexReader::rescore_hits`) mirroring the `combine_boost` error shape. The guard rejects the query at the earliest point the non-finite total is produced, before it can reach `ScoredTerm` or the scoring heap.

## Reproduction

Before the guard, a `bool.should` with two clauses carrying `1.8e38` boosts on the same `field:term` key accumulates to `3.6e38 → +Inf` and silently corrupts ranking or breaks serialisation. After the guard, the query is rejected at plan/expansion boundary with `accumulated boost for term \`body:rust\` overflows f32; reduce per-clause boost values`.

## Test plan

- [x] `cargo test --package searchlite-core --test multi_field` — 14 tests pass, including two new tests (`bool_rejects_accumulated_term_weight_overflow`, `bool_accepts_accumulated_term_weight_within_range`).
- [x] `cargo test --package searchlite-core` — full core test suite passes (770+ tests).
- [x] `cargo fmt --package searchlite-core --check` — clean.

Closes #401

### DIFF
--- a/searchlite-core/src/api/reader.rs
+++ b/searchlite-core/src/api/reader.rs
@@ -1272,6 +1272,16 @@ impl IndexReader {
         term.group_fields.clone(),
       ));
       entry.1 += term.weight;
+      // BUG-401: individual boosts are validated finite by combine_boost, but the
+      // sum of N finite f32 values can still overflow to +Inf when the same
+      // field:term key appears across multiple query clauses. Reject at the
+      // accumulation point so Inf never reaches ScoredTerm / score_tf.
+      if !entry.1.is_finite() {
+        bail!(
+          "accumulated boost for term `{}` overflows f32; reduce per-clause boost values",
+          term.key
+        );
+      }
       if entry.3.is_none() && term.group_fields.is_some() {
         entry.3 = term.group_fields.clone();
       }
@@ -2105,6 +2115,14 @@ impl IndexReader {
             .entry(term.key.clone())
             .or_insert((term.field.clone(), 0.0, term.leaf));
         entry.1 += term.weight;
+        // BUG-401: guard against finite-but-additive overflow into +Inf when
+        // multiple rescore clauses share the same field:term key.
+        if !entry.1.is_finite() {
+          bail!(
+            "accumulated boost for term `{}` overflows f32; reduce per-clause boost values",
+            term.key
+          );
+        }
       }
       // BM25 N must use total doc_count (including deleted documents) so it matches
       // the df basis (postings.len() includes deleted documents). See BUG-360.
--- a/searchlite-core/tests/multi_field.rs
+++ b/searchlite-core/tests/multi_field.rs
@@ -754,3 +754,92 @@ fn multi_match_ignores_overflow_boost_on_numeric_field_kind() {
     );
   }
 }
+
+#[test]
+fn bool_rejects_accumulated_term_weight_overflow() {
+  // BUG-401: `term_weights` in `IndexReader::search` sums per-term weights
+  // with `entry.1 += term.weight` when multiple query clauses resolve to the
+  // same `field:term` key. Each individual boost is validated finite by
+  // `combine_boost` at plan/expansion time (BUG-381 / BUG-396), but the sum
+  // of N finite f32 values can still overflow `f32::MAX` (~3.4e38) to `+Inf`.
+  // Before the guard, that `Inf` propagated into `ScoredTerm.weight` and
+  // through BM25 `score_tf` into the WAND/brute-force heap, corrupting
+  // ranking (WAND+Sum silently clamped to 0.0 per BUG-364) or producing an
+  // `Inf` in the serialised response (HTTP 500 on `serde_json::to_vec`).
+  // The guard must fire at accumulation time and mirror the
+  // `combine_boost` error shape.
+  let (_tmp, reader) = setup_reader();
+  // Two `should` clauses, each matching `body:rust` with a boost that is
+  // individually finite (f32::MAX ≈ 3.4e38). Their accumulated sum
+  // (3.6e38) exceeds f32::MAX and overflows to +Inf.
+  let query = QueryNode::Bool {
+    must: Vec::new(),
+    should: vec![
+      QueryNode::Term {
+        field: "body".into(),
+        value: "rust".into(),
+        boost: Some(1.8e38),
+      },
+      QueryNode::Term {
+        field: "body".into(),
+        value: "rust".into(),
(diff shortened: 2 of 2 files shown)


davidkelley:
@codex review

